PR #6, opened 2025-03-11
Add: Dashboard db migrations & tooling

ben-fornefeld:
this pr moves dashboard specific migrations inside this repo. also it adds tooling to work with db migrations.


scripts/apply-migrations.ts
```diff
@@ -0,0 +1,292 @@
...
+    await db(`
+      CREATE TABLE IF NOT EXISTS public._migrations (
+        filename TEXT PRIMARY KEY,
+        applied_at TIMESTAMP WITH TIME ZONE DEFAULT CURRENT_TIMESTAMP,
+        checksum TEXT
+      );
+    `)
+    await db(`
+      GRANT ALL ON public._migrations TO supabase_admin;
+    `)
```
jakubno, line 40: could you explain why you added?

scripts/apply-migrations.ts
```diff
@@ -0,0 +1,292 @@
...
+const db = new SQL(process.env.POSTGRES_CONNECTION_STRING)
+
+/**
+ * Creates or verifies the existence of the migrations tracking table
+ * This table keeps track of which migrations have been applied to prevent
+ * duplicate runs and provide an audit trail
+ */
+async function ensureMigrationsTable() {
+  try {
+    await db(`
```
jakubno, line 31: this isn't working because correct syntax is 
```suggestion
    await db`
```

scripts/apply-migrations.ts
```diff
@@ -0,0 +1,292 @@
...
+ * @param filename - The name of the migration file to check
+ * @param content - The content of the migration file for checksum verification
+ * @returns Object with applied status and whether checksum matches
+ */
+async function isMigrationApplied(
+  filename: string,
+  content: string
+): Promise<{ applied: boolean; checksumMatch: boolean }> {
+  // Query to get the existing migration record
+  const result = await db.unsafe(
```
jakubno, line 60: there's no need to this to be unsafe

scripts/apply-migrations.ts
```diff
@@ -0,0 +1,292 @@
...
+ * @param filename - The name of the migration file
+ * @param content - The content of the migration file (for checksum)
+ */
+async function recordMigration(
+  filename: string,
+  content: string
+): Promise<void> {
+  // Convert checksum to string to avoid bigint overflow
+  const checksum = String(Bun.hash(content))
+  await db.unsafe(
```
jakubno, line 100: no need to be unsafe

README.md
```diff
@@ -179,3 +186,92 @@ If you need help or have questions:
...
+This will:
+1. Create the migrations tracking table if it doesn't exist
+2. Apply any migrations that haven't been run yet
+3. Skip already-applied migrations
+4. Provide a summary of applied/skipped migrations
+
+### Migration Safety
+
+The system includes several safety features:
+- Migrations are applied in a transaction (all-or-nothing)
```
jakubno, line 266: not true!

jakubno (changes requested):
In readme you are missing bun as a prerequisite, also it needs to be 1.2 + because of SQL. I guess npm is not supported anymore.


Not sure if we want to replicate migration logic here, like splitting sql statements etc, why are we solving already solved problem.

AI is really good, but not checking the code afterwards or testing it ngmi..

ben-fornefeld:
> In readme you are missing bun as a prerequisite, also it needs to be 1.2 + because of SQL. I guess npm is not supported anymore.
i knew all of this. wanted to try the bun sql package. not the best decision for public use indeed.

> Not sure if we want to replicate migration logic here, like splitting sql statements etc, why are we solving already solved problem.
fair.

> AI is really good, but not checking the code afterwards or testing it ngmi..
code was of course checked & tested multiple times. i did bad decisions while doing so is a better way to put it.

jakubno:
> code was of course checked & tested multiple times. i did bad decisions while doing so is a better way to put it.

E.g. migrations weren't tested, at least not without some already existing tables, because it was failing as I mentioned in PR

ben-fornefeld:
> > code was of course checked & tested multiple times. i did bad decisions while doing so is a better way to put it.
> 
> E.g. migrations weren't tested, at least not without some already existing tables, because it was failing as I mentioned in PR

i tested it multiple times with multiple scenarios. maybe there was some edge case i did not caught. just want to mention that i of course test the code i write and not just push blindly on purpose.

jakubno:
I applied it to my database, I think going from clean state isn't an edge case

README.md
```diff
@@ -75,21 +75,28 @@ vercel storage add
...
+2. Run the migrations by running the following command:
+```bash
+bun run db:migrations:apply
+```
+
+#### d. Supabase Storage Setup
 1. Go to Storage > Buckets
 2. Create a new **public** bucket named `profile-pictures`
-3. Apply storage access policies by running the SQL from [supabase/policies/buckets.sql](supabase/policies/buckets.sql) in the Supabase SQL Editor:
+3. Apply storage access policies by running the SQL from [migrations/supabase/profile-picture-bucket.sql](migrations/supabase/profile-picture-bucket.sql) in the Supabase SQL Editor:
```
jakubno, line 88:
```suggestion
```

jakubno (changes requested):

jakubno (approved):

PR #12, opened 2025-03-17
Fix templates name column min size & truncation

ben-fornefeld:
this pr adds a minSize attribute to the `name` column in templates table to prevent to small column sizing & truncates the data when it overflows


jakubno (approved):
Adding before and after would be the best

PR #15, opened 2025-03-21
Add Zerobounce auth email validation

ben-fornefeld:
this pr integrates zerobounce email validation to facilitate higher quality sign-ups. it adds basic validation for the email address and also checks if it's marked as "alternate", which prompts the user to use his primary email. if the user tries to sign in again via the alternate, it allows a sign-up as usual.


mishushakov:
Not sure if this is a good idea to have a third-party collect our user's e-mails. Need to ask @mlejva

mishushakov (approved):

PR #19, opened 2025-03-26
Fix middleware rewrite handling to comply with CVE-2025-29927

ben-fornefeld:
Although mentioned in [CVE-2025-29927](https://github.com/advisories/GHSA-f82v-jwr5-mffw), the version upgrade to resolve the security vulnerability in NextJs in #18 was not enough to fix rewriting issues. This pr explored ways to get around the 406 blockage and in the end settled on:
- using `NextResponse.rewrite` instead of `fetch & new NextResponse`

This approach is far from ideal in terms of performance. It seems like the native NextJs rewrite is doing a lot under the hood before sending the html to the client, potentially adapting additional assets loaded by the rewritten site, as this latency seems to increase with more assets.

### DIFF
--- a/src/server/middleware.ts
+++ b/src/server/middleware.ts
@@ -220,12 +220,22 @@ export const handleUrlRewrites = async (
   }
 
   try {
-    const res = await fetch(url.toString(), { ...request })
+    const res = await fetch(url.toString(), {
+      headers: {
+        'User-Agent': 'Mozilla/5.0 (compatible; E2BProxy/1.0)',
+        Accept:
+          'text/html,application/xhtml+xml,application/xml;q=0.9,image/webp,*/*;q=0.8',
+        'Accept-Language': 'en-US,en;q=0.5',
+        'Accept-Encoding': 'gzip, deflate, br',
+        Connection: 'keep-alive',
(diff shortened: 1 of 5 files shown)


src/server/middleware.ts
```diff
@@ -2,12 +2,13 @@ import { checkUserTeamAuthorization, resolveTeamId } from '@/lib/utils/server'
...
 import { KV_KEYS } from '@/configs/keys'
 import { NextRequest, NextResponse } from 'next/server'
-import { replaceUrls } from '@/configs/domains'
+import { LANDING_PAGE_DOMAIN, replaceUrls } from '@/configs/domains'
 import { COOKIE_KEYS } from '@/configs/keys'
 import { AUTH_URLS, PROTECTED_URLS } from '@/configs/urls'
 import { supabaseAdmin } from '@/lib/clients/supabase/admin'
 import { z } from 'zod'
 import { createServerClient } from '@supabase/ssr'
+import { logDebug } from '@/lib/clients/logger'
```
jakubno, line 11:
```suggestion
```

jakubno (approved):

PR #24, opened 2025-03-31
Change infra and billing api authorization headers

ben-fornefeld:
To prepare for api key & access token hashing, we move away from authorizing against our infrastructure and billing api via these, since we cannot retrieve / store them for the user anymore soon. 

This pr provides authorization headers to these apis via auth jwts and/or teamId instead.

### DIFF
--- a/src/server/billing/get-billing-limits.ts
+++ b/src/server/billing/get-billing-limits.ts
@@ -22,7 +22,7 @@ export const getBillingLimits = authActionClient
         method: 'GET',
         headers: {
           'Content-Type': 'application/json',
-          ...SUPABASE_AUTH_HEADERS(session.access_token, teamId),
+          ...SUPABASE_AUTH_HEADERS(session.access_token),
         },
       }
     )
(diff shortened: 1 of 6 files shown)


src/server/billing/get-billing-limits.ts
```diff
@@ -15,17 +14,15 @@ export const getBillingLimits = authActionClient
...
+    const { session } = ctx
 
     const res = await fetch(
       `${process.env.BILLING_API_URL}/teams/${teamId}/billing-limits`,
       {
         method: 'GET',
         headers: {
           'Content-Type': 'application/json',
-          [USER_ACCESS_TOKEN_HEADER]: accessToken,
+          ...SUPABASE_AUTH_HEADERS(session.access_token, teamId),
```
dobrac, line 25: just a note, you don't need to pass the teamId to places, where access token was previously required

ben-fornefeld: yes, i forgot to not pass it in the helper in some routes where it is still present. but good point i will fix that quickly

dobrac (approved):

PR #34, opened 2025-04-04
Refactor content page rewrites

ben-fornefeld:
After #18 we were using NextResponse.rewrite to rewrite our content pages inside `middleware.ts`. The exact reasons are unclear but this had some massive latency overhead on every rewritten route which was not ideal. This pr adds a catch-all route handler which catches all paths that are not defined by the next router. Here we do our content page rewrites & add cache policies to the fetches & responses with revalidation after 15 minutes. Since we use a catch-all handler, we need to manually handle our root 404 page as well. Returning nextjs's `notFound()` did only send a 404 status to the browser, but we want to render our custom not-found page. Hence I added a new route just holding our not-found component and rewrite accordingly in the catch-all handler.

This results in much faster performance of our content pages & separates the rewriting process from the middlewares business logic for clearer separation of concerns.


ValentaTomas (approved):
Looks good to me—we went over the details together.

PR #37, opened 2025-04-08
Fix table sorting function to not sort based on date time string

ben-fornefeld:
Currently the table date time columns sorted based on their date time string representations. This pr fixes this, along with disabling multiple sortings at once, since this behavior was confusing users.

### DIFF
--- a/src/features/dashboard/templates/table-config.tsx
+++ b/src/features/dashboard/templates/table-config.tsx
@@ -300,11 +300,24 @@ export const useColumns = (deps: unknown[]) => {
         header: 'Created At',
         size: 250,
         minSize: 140,
-        cell: ({ getValue }) => (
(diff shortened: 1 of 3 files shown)


src/features/dashboard/templates/table-config.tsx
```diff
@@ -300,11 +300,24 @@ export const useColumns = (deps: unknown[]) => {
         header: 'Created At',
         size: 250,
         minSize: 140,
-        cell: ({ getValue }) => (
-          <div className="text-fg-500 truncate font-mono text-xs">
-            {getValue() as string}
-          </div>
-        ),
+        cell: ({ row, getValue }) => {
```
jakubno, line 303: isn't it possible to sort on preprocessed values?

jakubno: sorry I mixed sort and cell

PR #41, opened 2025-04-12
Refactor team creation to not depend on billing

ben-fornefeld:
Because the dashboard should be self-host able without the need to include our billing backend, this pr:
1. refactors team creation action to create team via direct db access
2. adds proper team name validation on client & server side
3. improves optimistic team name updates to eliminate refetch state race condition


mlejva (approved):

src/server/team/types.ts
```diff
@@ -11,3 +12,52 @@ export type TeamMember = {
...
+ * - "   " -> Fails with "Team name cannot be empty" (becomes empty after trim)
+ * - "Team@Work" -> Fails with "Team name can only contain letters, numbers, and spaces"
+ * - "Dev-Ops" -> Fails with "Team name can only contain letters, numbers, and spaces"
+ * - "Very Long Team Name That Exceeds The Character Limit" -> Fails with "Team name cannot be longer than 32 characters"
+ *
+ * Validation errors:
+ * - Empty string or strings that become empty after trim: "Team name cannot be empty"
+ * - Strings longer than 32 characters: "Team name cannot be longer than 32 characters"
+ * - Strings with special characters: "Team name can only contain letters, numbers, and spaces"
+ */
```
mishushakov, line 42: The comment is maybe too large, one-liner would have been enough to explain the regex

PR #42, opened 2025-04-13
Add NO_INDEX & VERBOSE flags

ben-fornefeld:
In order to prevent search engine indexing for our staging / juliett dashboards, but keep rewriting intact for each, this pr adds a NO_INDEX environment variable which takes precautions to tell site crawlers to not index this site.

Additionally this pr:
- adds a NEXT_PUBLIC_VERBOSE environment variable, for better observability in production builds if needed
- try-catches json serialization for sandboxes metadata search

### DIFF
--- a/src/app/(rewrites)/[[...slug]]/route.ts
+++ b/src/app/(rewrites)/[[...slug]]/route.ts
@@ -81,23 +81,9 @@ export async function GET(request: NextRequest): Promise<Response> {
       const newHeaders = new Headers(res.headers)
       newHeaders.delete('content-encoding')
 
-      // take precautions if NO_INDEX is set
+      // Add noindex header if NO_INDEX is set
       if (NO_INDEX) {
-        // add noindex header
         newHeaders.set('X-Robots-Tag', 'noindex, nofollow')
-        // remove any existing robots meta tags and add our own
-        const modifiedHtml = modifiedHtmlBody
-          .replace(/<meta[^>]*name=["']robots["'][^>]*>/gi, '')
-          .replace(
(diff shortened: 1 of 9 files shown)


src/app/(rewrites)/[[...slug]]/route.ts
```diff
@@ -80,6 +81,25 @@ export async function GET(request: NextRequest): Promise<Response> {
...
+        // add noindex header
+        newHeaders.set('X-Robots-Tag', 'noindex, nofollow')
+        // remove any existing robots meta tags and add our own
+        const modifiedHtml = modifiedHtmlBody
+          .replace(/<meta[^>]*name=["']robots["'][^>]*>/gi, '')
+          .replace(
+            /<head[^>]*>/i,
+            (match) =>
+              `${match}<meta name="robots" content="noindex, nofollow">`
+          )
```
Comment (mishushakov) [inline src/app/(rewrites)/[[...slug]]/route.ts:95]: Is there more elegant way to do this instead of rewriting HTML

Comment (ben-fornefeld) [reply src/app/(rewrites)/[[...slug]]/route.ts:95]: not really afaik. the header alone is not enough. but this is pretty stable, i have checked it on all our possible content rewrite domains.

Comment (mishushakov) [reply src/app/(rewrites)/[[...slug]]/route.ts:95]: but you can do it dynamically?
https://nextjs.org/docs/app/api-reference/file-conventions/metadata/robots#generate-a-robots-file

Comment (ben-fornefeld) [reply src/app/(rewrites)/[[...slug]]/route.ts:95]: yes, very good call with `robots.ts`, since it's authoritative than meta tags

mishushakov (approved):

PR #48, opened 2025-04-21
Prevent docs app interactivity issues by disabling caching for /docs rewrites

ben-fornefeld:
This PR addresses an issue where cached HTML continues to reference outdated JS chunks after docs app redeployments, breaking interactivity. The solution:

1. Creates a centralized config file to manage rewrite rules
2. Handles native rewrites in middleware to disable caching for /docs routes
3. Ensures route handlers use the same config

The change eliminates the 15-minute cache window where outdated HTML would request non-existent asset files, preventing broken user experiences after deployments of our docs application.

Additionally this pr:
- replaces replaceUrls function with cloudflares HTMLRewriter api to remove invalid absolute urls from html
- rewrites /ai-agents in middleware


src/lib/utils/rewrites.ts
```diff
@@ -0,0 +1,59 @@
...
+
+  return { config: null, rule: null }
+}
+
+function replaceUrls(
+  text: string,
+  prefix: string = '',
+  suffix: string = ''
+): string {
+  const pattern = suffix
```
mishushakov, line 47: what does the regex match here?

ben-fornefeld: was taken from the old dashboard repo but got removed in favor of HTMLRewriter now

src/app/(rewrites)/[[...slug]]/route.ts
```diff
@@ -76,26 +56,71 @@ export async function GET(request: NextRequest): Promise<Response> {
...
+                    'href',
+                    url.pathname + url.search + url.hash
+                  )
+                } catch (e) {
+                  // Ignore invalid URLs
+                  console.error('Failed to parse URL for rewriting:', href, e)
+                }
+              }
+            },
+          })
```
Comment (mishushakov) [inline src/app/(rewrites)/[[...slug]]/route.ts:103]: de-duplicate this please

mishushakov (changes requested):

mishushakov (approved):

PR #52, opened 2025-04-24
Remove ai-agents middleware rewrite

ben-fornefeld:
This pr removes references to our  framer landing page by removing `LANDING_PAGE_FRAMER_DOMAIN` from global/rewrite config and sitemap generation.

Additionally:
- it adds a `seo` config for marking path globs as deprecated routes to inform seo crawlers about deprecation to get them removed from indexing faster by returning `410` (no content) status codes instead of `404` in catch-all route handler
- configures `/ai-agents/**` paths to return `410`


src/app/(rewrites)/[[...slug]]/route.ts
```diff
@@ -15,6 +16,22 @@ const REVALIDATE_TIME = 900 // 15 minutes ttl
...
+      const regex = new RegExp(pattern)
+      if (regex.test(pathname)) {
+        return new Response(null, { status: 410 })
+      }
+    } catch (error) {
+      logWarning(WARNING_CODES.SEO, pattern, error)
+
+      continue
+    }
+  }
```
Comment (mishushakov) [inline src/app/(rewrites)/[[...slug]]/route.ts:33]: can't you just

```
DEPRECATED_PATH_REGEXPS.some(regex => regex.test(pathname))
```

src/configs/seo.ts
```diff
@@ -0,0 +1,14 @@
...
+ * Configuration for SEO-related settings.
+ */
+
+/**
+ * An array of regular expression strings. Paths matching any of these regexps
+ * will return a 410 (Gone) status.
+ * Ensure these are valid JavaScript RegExp patterns.
+ * Example: ['^/old-blog(/.*)?$', '^/features/removed-feature$']
+ */
+export const DEPRECATED_PATH_REGEXPS: string[] = [
```
mishushakov, line 11: deprecatedPaths

src/configs/seo.ts
```diff
@@ -0,0 +1,14 @@
...
+  '^/ai-agents(/.*)?$', // Matches /ai-agents and all subpaths
```
mishushakov, line 12: /^\/ai-agents(\/.*)?$/

src/configs/seo.ts
mishushakov: string[] > RegExp[]

mishushakov (changes requested):

src/app/(rewrites)/[[...slug]]/route.ts
Comment (ben-fornefeld) [reply src/app/(rewrites)/[[...slug]]/route.ts:33]: i didn't use native `RegExp` for configuration to make this easier/more readable when configuring

ben-fornefeld:
@mishushakov the way you suggested makes it slightly annoying to configure. remember when using js native regexp, for every literal `/` match you have to prefix with a `\`. since we try to match paths here, that can get quite cumbersome.

src/app/(rewrites)/[[...slug]]/route.ts
Comment (ben-fornefeld) [reply src/app/(rewrites)/[[...slug]]/route.ts:33]: this way we do not get invalid regex patterns logged.

ben-fornefeld:
refactored to glob pattern matching for easier configurability

src/configs/seo.ts
```diff
@@ -0,0 +1,17 @@
...
+/**
+ * An array of glob patterns for deprecated routes that will return a 410 (Gone) status.
+ *
+ * Pattern syntax examples:
+ * - '/ai-agents' - Exact match for /ai-agents
+ * - '/ai-agents/*' - Matches a single segment (e.g., /ai-agents/foo)
+ * - '/ai-agents/**' - Matches multiple segments (e.g., /ai-agents/foo/bar)
+ * - '/ai-agents/{foo,bar}' - Matches /ai-agents/foo or /ai-agents/bar
+ * - '/ai-agents/!(foo|bar)' - Matches anything except /ai-agents/foo or /ai-agents/bar
+ */
```
mishushakov, line 14: remove comments, I think the code is self-explanatory

src/lib/utils/seo.ts
```diff
@@ -0,0 +1,43 @@
...
+ *
+ * @param pathname URL pathname to check against deprecated patterns
+ * @returns boolean indicating if the path is deprecated
+ */
+export function isPathDeprecated(pathname: string): boolean {
+  // Normalize the pathname to ensure consistent matching
+  const normalizedPath =
+    pathname.endsWith('/') && pathname !== '/'
+      ? pathname.slice(0, -1) // Remove trailing slash except for root path
+      : pathname
```
mishushakov, line 17: do you actually need to normalize it?

src/configs/seo.ts
ben-fornefeld: it's not. if i just want to configure new paths and don't know anything about the repo this is helpful

PR #62, opened 2025-05-13
Improve usage page prose

ben-fornefeld:
This pr improves the sandboxes chart code to be more readable + adds an indication for the current month (billing period) to be easily comparable with the other usage charts. Also this pr improves the general prose of the usage page to be more in sync with what the components show.

### DIFF
--- a/src/features/dashboard/usage/cost-card.tsx
+++ b/src/features/dashboard/usage/cost-card.tsx
@@ -67,7 +67,7 @@ export function CostCard({
       <CardHeader>
         <CardTitle className="font-mono">Usage Costs</CardTitle>
         <CardDescription>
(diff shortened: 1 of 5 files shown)


src/features/dashboard/usage/cost-card.tsx
```diff
@@ -67,7 +67,7 @@ export function CostCard({
       <CardHeader>
         <CardTitle className="font-mono">Usage Costs</CardTitle>
         <CardDescription>
-          Total cost of all resources this month.
+          Total cost of all resources each month.
```
dobrac, line 70: shouldn't we use "per month" instead?

dobrac (approved):

PR #72, opened 2025-06-02
 Refactor: HTML Rewriting and SEO Tag Injection based on NO_INDEX

ben-fornefeld:
This PR refactors the HTML rewriting logic for proxied content and introduces a new approach for injecting SEO tags (robots meta and canonical links). This is reasonable, because we have multiple versions of the dashboard hosted on the internet and want to make sure only the prod proxied pages stay indexed via an env flag `NO_INDEX`.

**Key Changes:**

*   **Replaced `HTMLRewriter` with `Cheerio`:**
    *   The previous implementation used `@worker-tools/html-rewriter` for modifying HTML. This has been replaced with `cheerio` since the "prepend" operation did not work with `HTMLRewriter`.
    *   This change affects how absolute URLs are rewritten to relative paths in proxied content.
*   **Centralized SEO Tag Injection:**
    *   A new utility function `rewriteContentPagesHtml` (in `src/lib/utils/rewrites.ts`) now handles both rewriting absolute hrefs and injecting SEO tags.
    *   This function uses `cheerio` to:
        *   Remove any existing `<meta name="robots">` and `<link rel="canonical">` tags.
        *   Prepend new `robots` meta tags (`noindex,nofollow,noarchive` or `index,follow` based on the `NO_INDEX` flag).
        *   Prepend new `canonical` link tags, pointing to the `e2b.dev` version of the page.
*   **Updated Route Handler (`src/app/(rewrites)/[[...slug]]/route.ts`):**
    *   The GET handler now uses `rewriteContentPagesHtml` to process HTML responses from the proxied domain.
    *   It correctly sets SEO tags and rewrites links before returning the response.
*   **Added `robots: noindex, nofollow` to Dashboard Layout:**
    *   The `src/app/dashboard/layout.tsx` now includes metadata to set `robots` to `noindex, nofollow` for all dashboard pages.


src/lib/utils/rewrites.ts
```diff
@@ -36,4 +42,106 @@ function getRewriteForPath(
...
+ * It finds <a> and <link> tags whose href starts with the given prefix
+ * and replaces the href with a relative path (pathname + search + hash).
+ *
+ * @param $ The Cheerio API instance.
+ * @param rewrittenPrefix The absolute URL prefix (e.g., "https://example.com") to replace.
+ */
+function rewriteAbsoluteHrefsInDoc(
+  $: cheerio.CheerioAPI,
+  rewrittenPrefixes: string[]
+): void {
```
dobrac, line 97: is this still necessary? Is it fast enough?

dobrac (changes requested):

src/lib/utils/rewrites.ts
ben-fornefeld: It should be fine, since the routes that use these are data cached.

dobrac: But you don't need to prepend anymore the headers, as they're set on the source, do you?

ben-fornefeld: i do, for our static content pages.

dobrac: Oh, I see, by doc you don't mean docs, but document 😀

ben-fornefeld: the mechanism we use for our docs de/indexing would probably be much harder, if even not possible on our webflow landing page. since this is not really a concern here because of the caching, i kept it like this.

dobrac (approved):

PR #97, opened 2025-07-18
Improve: Cookie TTL's

ben-fornefeld:
This pr improves the following cookies to store UI state, such as the sidebar state:
- Sidebar state from `7 days` to `1 year`
- Selected-Team states from `7 days` to `1 year`

### DIFF
--- a/src/app/api/team/state/route.ts
+++ b/src/app/api/team/state/route.ts
@@ -8,16 +8,16 @@ const TeamStateSchema = z.object({
   teamSlug: z.string(),
 })
 
+const COOKIE_SETTINGS: Partial<ResponseCookie> = {
+  path: '/',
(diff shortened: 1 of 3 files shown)


src/app/api/team/state/route.ts
```diff
@@ -8,17 +8,17 @@ const TeamStateSchema = z.object({
   teamSlug: z.string(),
 })
 
+const COOKIE_SETTINGS: Partial<ResponseCookie> = {
+  path: '/',
+  maxAge: 60 * 60 * 24 * 365, // 1 year
```
0div, line 13: if this one is used for secrets (which it doesn't seem like at glance), i would keep it for 30 days, else carry on.

0div (approved):

PR #102, opened 2025-07-25
Fix: next-logger pino requirement

ben-fornefeld:
next-logger requires `pino` installed when used on the server, although we are using `winston` as the actual logger. this is a hot-fix and i will investigate in the future if we can mitigate here.

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -61,6 +61,7 @@
         "next-safe-action": "^7.10.4",
         "next-themes": "^0.4.4",
         "openapi-fetch": "^0.14.0",
+        "pino": "^9.7.0",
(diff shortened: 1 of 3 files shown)


mishushakov:
- can we use pino instead of winston?
- are we actually using edge functions anywhere? if not, feel free to remove this

ben-fornefeld:
> * can we use pino instead of winston?
> * are we actually using edge functions anywhere? if not, feel free to remove this

1. tried pino extensively, winston is better.
2. answered already in different pr, the edge runtime is still in use

mishushakov (approved):

PR #113, opened 2025-08-01
Feature: Add Disk Usage Metrics and Display to Sandbox List and UI

ben-fornefeld:
## Summary

This pull request introduces disk usage metrics to the sandbox management dashboard. The changes include backend support for tracking disk usage per sandbox, updates to mock data and type definitions, and new UI elements to display disk usage in the dashboard.

## Key Changes

- **API & Types**:
  - Added `diskUsed` and `diskTotal` fields to the API schemas (`openapi.yaml`).
  - Extended `ClientSandboxMetric` types and related type definitions to include `diskUsedGb` and `diskTotalGb`.
  - Updated transformation utilities to process disk metrics from API responses.

- **Mock Data**:
  - Enhanced mock generator to provide disk usage data for different sandbox templates.

- **UI/Frontend**:
  - Updated `ResourceUsage` component to support and display disk metrics.
  - Added disk usage to the sandbox details header via `ResourceUsageClient`.
  - Added a new `DiskUsageCell` and integrated it into the sandboxes table as a sortable column.
  - Updated all relevant components and hooks to handle the new disk metric fields.

## Impact

- Users can now view disk usage per sandbox both in the details header and as a column in the sandboxes table.
- Improves monitoring and resource management for sandboxes, especially for environments that rely on disk-intensive operations.

## Sandbox List
Screenshot 2025-08-01 at 2 32 44 PM

## Sandbox Details
Screenshot 2025-08-01 at 2 34 56 PM


ben-fornefeld:
> I would add some info (maybe ⚠️) if the envd version is lower than 0.2.4, because the metrics for disk won't be available.

@jakubno unfortunately i don't have envdVersion info on ListedSandbox afaik.

PR #136, opened 2025-09-10
UI tweaks and fixes to the Sandbox monitoring feature

vojtabohm:
- Limited label size, fixed color usage, and corrected alignment.
- Adjusted live badge: lighter weight and smaller dot.
- Reduced thickness of value-prose-big.
- Fixed Math.max on empty array by adding default 0.
- Corrected alignment of the third card so the live badge doesn’t affect layout.
- Optically centered the header cards.
- Baseline-aligned concurrent label with live badge.
- Reduced font size of graph time switcher.
- Fixed graph padding (removed extra px-1).
- Corrected spacing and alignment of the live container at the top.
- Increased tab font weight.

### DIFF
--- a/src/features/dashboard/sandboxes/live-counter.tsx
+++ b/src/features/dashboard/sandboxes/live-counter.tsx
@@ -17,15 +17,15 @@ export function LiveSandboxCounter({
   return (
     <div
       className={cn(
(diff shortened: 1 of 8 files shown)


vojtabohm:
@ben-fornefeld Glad to have helped! Thanks.

PR #140, opened 2025-09-12
Feature: Add "Last Used" column to API Keys page

ben-fornefeld:
This pr adds a new column, which informs users when an api key was last used. if the key creation date is before the production collection date and last_used is undefined, we show "N/A", since we can't know if it was used or not before the collection date.

### New API Keys Table
Screenshot 2025-09-12 at 3 15 42 PM


src/configs/versioning.ts
```diff
@@ -1,3 +1,5 @@
 export const SANDBOX_INSPECT_MINIMUM_ENVD_VERSION = '0.2.9'
 
 export const SANDBOXES_DISK_METRICS_MINIMUM_ENVD_VERSION = '0.2.4'
+
+export const API_KEYS_LAST_USED_FIRST_COLLECTION_DATE = new Date('2025-09-10')
```
jakubno, line 5: you don't need this, it's either null or last used date

jakubno (approved):

PR #147, opened 2025-09-23
Monitoring dashboard design tweaks

vojtabohm:
1. Copy on all labels made consistent
2. Moved the live label to the right
3. Unified the number formatting on graphs & cards
4. Reduced the height of cards
5. Switched the order of the two base cards (concurrent vs start rate) to make it consistent with the graphs below

Before:
Arc - 2025-09-24 at 12 13 45@2x

After:
Arc - 2025-09-24 at 12 12 14@2x

---

> [!NOTE]
> Updates monitoring dashboard UI with consistent labels/number formatting, right-aligned live badges, reduced card height, swapped card order, and tweaked chart/header layouts.
> 
> - **Monitoring Header**:
>   - **Card order**: Swap positions so `ConcurrentSandboxes` appears before `SandboxesStartRate`.
>   - **Copy**: Update subtitles to "(5-sec avg)" and rename to `Peak Concurrent Sandboxes` with "(30-day max)".
>   - **Layout**: Reduce card height (`min-h` tweaks) and adjust LIMIT label positioning for responsiveness.
> - **Charts**:
>   - **Concurrent chart (`concurrent-chart.client.tsx`)**:
>     - Header rewritten: explicit "Concurrent sandboxes" with `ReactiveLiveBadge` aligned to the right.
>     - Display value uses `formatAxisNumber` (was `formatDecimal(..., 1)`); header/time controls layout simplified and made responsive.
>   - **Start rate chart (`start-rate-chart.client.tsx`)**:
>     - Header aligns with concurrent chart: label + right-aligned `ReactiveLiveBadge`; spacing simplified.
> - **Number formatting**:
>   - Use `formatDecimal(..., 3)` for start rate in header client and adjust margins (`mt-1`), keeping values consistent across charts and cards.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 9e5fb3e8cd46d5e659e27d27044c9faac20dd2c3. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/src/features/dashboard/sandboxes/monitoring/header.tsx
+++ b/src/features/dashboard/sandboxes/monitoring/header.tsx
@@ -17,7 +17,7 @@ import {
 
 function BaseCard({ children }: { children: React.ReactNode }) {
   return (
-    <div className="p-4 md:p-6 max-md:not-last:border-b md:not-last:border-r h-full flex-1 w-full flex flex-col justify-center items-center gap-2 md:gap-3 relative max-md:min-h-[100px] md:min-h-[200px]">
+    <div className="p-4 md:p-6 max-md:not-last:border-b md:not-last:border-r flex-1 w-full flex flex-col justify-center items-center gap-2 md:gap-3 relative min-h-[100px] md:h-45">
       {children}
     </div>
   )
@@ -50,7 +50,7 @@ export default function SandboxesMonitoringHeader({
   params: Promise<SandboxesMonitoringPageParams>
 }) {
   return (
-    <div className="flex md:flex-row flex-col items-center border-b w-full md:min-h-52 max-md:py-2">
+    <div className="flex md:flex-row flex-col items-center border-b w-full max-md:py-2">
       <BaseCard>
         <SemiLiveBadge className="absolute left-3 top-3 md:left-6 md:top-6" />
         <Suspense fallback={<Skeleton className="w-16 h-8" />}>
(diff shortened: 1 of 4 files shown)


src/features/dashboard/sandboxes/monitoring/header.tsx
```diff
@@ -84,7 +83,7 @@ export default function SandboxesMonitoringHeader({
           Max<span className="max-md:hidden"> Concurrent Sandboxes</span>
           <span className="md:hidden"> Concurrent</span>
           <br className="max-md:hidden" />
-          <span className="max-md:hidden">(Last 30 Days)</span>
+          <span className="max-md:hidden">(30-day max)</span>
         </BaseSubtitle>
```
ben-fornefeld, line 87: i like "AVG" here better visually, but shouldn't we be consistent between 'avg' / 'average'?

i would suggest, if we go with AVG, we should at least also change the "30 seconds average" text in the tooltip for now, to make it consistent everywhere except for the big value above the chart, since median is also fully written out

ben-fornefeld (changes requested):
there is still a few nit's here on the mobile view:
- start rate card repeats itself text wise
- "concurrent" should probably also be 'concurrent sandboxes' here, since you changed it for the chart?
- the limit in the third card seems off 

Image

these issues might have been here before already, but since we're on it, can we fix it quickly?

vojtabohm:
Fixed
Arc - 2025-10-03 at 15 29 06@2x

ben-fornefeld:
bugbot run

ben-fornefeld (approved):
lgtm, thanks @vojtabohm !

PR #260, opened 2026-02-27
Fix: Prevent sensitive data from leaking into logs

taheerahmed:
## Summary

Two fixes to prevent sensitive data from appearing in application logs.

### 1. Auth code removed from callback error log

**File:** `src/app/api/auth/callback/route.ts` (line 42)

```diff
 context: {
-  code,
+  hasCode: !!code,
   origin,
   returnTo,
   redirectTo,
 },
```

The full auth code was logged when `exchangeCodeForSession` failed. Changed to a boolean, consistent with how the same route already logs it at line 23:

```typescript
// Line 23 — already uses boolean (correct)
context: {
  code: !!code,
  ...
}
```

### 2. Missing fields added to pino redaction paths

**File:** `src/lib/clients/logger/logger.ts` (lines 54-76)

Added `access_token` and `signatureSecret` at all 3 nesting levels.

**Why these were missed:** Pino redaction uses **exact field name matching**, not substring matching. The existing paths did not cover these because:

| Existing path | Catches | Does NOT catch |
|---|---|---|
| `accessToken` | `{ accessToken: "..." }` | `{ access_token: "..." }` |
| `*.secret` | `{ input: { secret: "..." } }` | `{ input: { signatureSecret: "..." } }` |

Supabase sessions use `access_token` (snake_case), and webhook actions use `signatureSecret` — both bypassed the existing redaction config.

## What this does NOT change

The `clientInput` logging in `src/lib/clients/action.ts:80` still logs full action inputs, but the newly added redaction paths now cover `signatureSecret` flowing through it. Whether to remove or allowlist `clientInput` logging entirely is a broader design decision best left to the maintainers.

Closes #259


drankou:
Addressed in #329

PR #292, opened 2026-04-10
Refactor: Team settings page

sarimrmalik:
[Ticket](https://linear.app/e2b/issue/ENG-3762/refactor-team-settings-page-according-to-figma)

- Refactors the team settings page into simpler, focused sections for avatar, name, and team info, including inline team-name editing with auto-fit text and clearer validation feedback.
- Reworks the team members experience around a richer client-side table with search, cleaner card/layout structure, and improved add/remove member flows.
- Moves more team-management logic into the tRPC layer and shared schemas, removing older action-based paths and tightening repository-backed validation and error handling.


vojtabohm:
Looks great! Just two small things: the overall section spacing from the top edge and the line height of the main font (should be 32, is 40).

Figma - 2026-04-17 at 23 49 18@2x

src/core/shared/schemas/file.ts
ben-fornefeld, line 1: can you check if there is a better, more native way, to do file uploads inside trpc than base64 ?

src/features/dashboard/members/member-card.tsx
```diff
@@ -1,51 +1,34 @@
...
-const MembersPageContentLoader = async ({ params }: MemberCardProps) => {
-  const { teamSlug } = await params
+export const MemberCard = () => {
+  const { team } = useDashboard()
+  const trpc = useTRPC()
+  const {
+    data: members,
+    error,
+    isLoading,
+  } = useQuery(trpc.teams.members.queryOptions({ teamSlug: team.slug }))
```
ben-fornefeld: let's either: 
- `useSuspenseQuery` and handle loading / error state via error/suspense boundaries in the components that actually need the data
- increase the `staleTime` on prefetch and useQuery, such that it does not double fetch on mount (one server fetch, one client side mount fetch)

src/features/dashboard/settings/general/team-name.tsx
```diff
@@ -0,0 +1,211 @@
...
+  useToast,
+} from '@/lib/hooks/use-toast'
+import { getTRPCValidationMessages } from '@/lib/utils/trpc-errors'
+import { useTRPC } from '@/trpc/client'
+import { Button } from '@/ui/primitives/button'
+import { CheckIcon, EditIcon } from '@/ui/primitives/icons'
+
+const TEAM_NAME_MAX_FONT_SIZE_PX = 32
+const TEAM_NAME_MIN_FONT_SIZE_PX = 18
+const TEAM_NAME_LINE_HEIGHT_PX = 32
```
ben-fornefeld: why are we doing this?

src/core/server/api/routers/teams.ts
```diff
@@ -1,23 +1,217 @@
...
     if (!teamsResult.ok) {
       throwTRPCErrorFromRepoError(teamsResult.error)
     }
 
     return teamsResult.data
   }),
+  create: authedTeamsRepositoryProcedure
+    .input(CreateTeamSchema)
+    .mutation(async ({ ctx, input }) => {
+      const result = await ctx.teamsRepository.createTeam(input.name)
```
ben-fornefeld: for create team, we should use `users-teams.repository.server.ts`, same as we did for the `listUsersTeams()`. this way you can also remove `authedTeamsRepositoryProcedure`

src/features/dashboard/sidebar/create-team-dialog.tsx
```diff
@@ -35,48 +37,58 @@ interface CreateTeamDialogProps {
...
 
-        if (result.data && result.data.slug) {
-          router.push(PROTECTED_URLS.SANDBOXES(result.data.slug))
-          router.refresh()
-        }
+        if (!team.slug) return
 
-        handleDialogChange(false)
+        router.push(PROTECTED_URLS.SANDBOXES(team.slug))
+        router.refresh()
```
ben-fornefeld: can you check if we still need `router.refresh()` here?

src/features/dashboard/sidebar/menu.tsx
```diff
@@ -49,23 +49,17 @@ export default function DashboardSidebarMenu({
...
-              >
-                <AvatarImage
-                  src={team.profilePictureUrl || undefined}
-                  className="group-data-[collapsible=icon]:size-full object-cover object-center"
-                />
-                <AvatarFallback className="bg-bg-hover border-0">
-                  {team.name?.charAt(0).toUpperCase() || '?'}
-                </AvatarFallback>
-              </Avatar>
+                imageClassName="group-data-[collapsible=icon]:size-full"
```
ben-fornefeld: let's have a const object `classNames` if a component takes multiple classNames. 

```typescript
classNames={{
  image: "group-data-[collapsible=icon]:size-full"
}}
```

src/lib/utils/trpc-errors.ts
```diff
@@ -1,11 +1,26 @@
...
+      zodError: z
+        .object({
+          formErrors: z.array(z.string()),
+          fieldErrors: z.record(z.string(), z.array(z.string()).optional()),
+        })
+        .nullable()
+        .optional(),
+    })
+    .optional(),
+})
```
ben-fornefeld: why are we doing this?

src/ui/primitives/loader.tsx
```diff
@@ -1,3 +1,5 @@
+'use client'
```
ben-fornefeld: this might cause the loader to be stuck on first server render, before client side js is executed. can you verify this and if true, refactor the loader implementation such that it is able to be server side rendered? i think we can achieve this with native css / react css modules

ben-fornefeld (changes requested):

src/features/dashboard/members/member-card.tsx
sarimrmalik: good callout, this would trigger a double fetch in the current state

src/features/dashboard/settings/general/team-name.tsx
sarimrmalik: this is the design spec from Vojtech, the team name should auto-shrink to fit the available width rather than truncate or wrap. see [Figma](https://www.figma.com/design/PgyPoTwMj8KkqOgUbJ18OL/Dashboard-2.0?node-id=1819-115534&m=dev). 

the three constants are the inputs to the measure-and-shrink loop in the effect below: starting size, floor, and a fixed line-height to keep the row from jumping as the font size changes.

src/core/server/api/routers/teams.ts
sarimrmalik: good idea, fixed

src/features/dashboard/sidebar/create-team-dialog.tsx
sarimrmalik: not needed, removed

src/features/dashboard/sidebar/menu.tsx
sarimrmalik: i really like that, fixed

src/lib/utils/trpc-errors.ts
sarimrmalik: we're parsing the tRPC error envelope to pull out the server-side `ZodError` payload (`formErrors` + `fieldErrors`) so we can surface validation messages as toasts in `team-name.tsx` and `create-team-dialog.tsx.` 

Using a Zod schema + `safeParse` because the helper is called from many call sites whose error types don't unify (each procedure's TRPCClientErrorLike is parameterized differently, plus `useQuery` returns `Error | null`), and tRPC's typed `fieldErrors` collapses to `{}` since `flattenError` is over `unknown`, so we'd need a cast either way.

src/ui/primitives/loader.tsx
sarimrmalik: good flag, fixed

src/core/shared/schemas/file.ts
sarimrmalik: good callout. i initially leaned into `base64` here because our existing support file upload flow already uses the same pattern, so I was keeping avatar upload consistent with that.

i did look into the more native tRPC v11 path: it supports `FormData` / non-JSON inputs directly, but because we currently use the batched streaming link for tRPC, uploads need to be routed through a separate non-batched `httpLink` transport. So it’s doable, but it adds a bit of client/server plumbing versus the simpler base64 shape.

ben-fornefeld: i see, thanks for the exploration!

src/lib/utils/trpc-errors.ts
ben-fornefeld: stepping back — i don't think we should need this helper at all. the standard trpc + zod pattern is to share
the input schema between procedure and form, validate client-side with `zodResolver`, and treat server
`zodError` as a backstop for schema drift. that way field errors render inline (not as toasts) and the
client never has to parse the error envelope.

looking at the call sites:
- `create-team-dialog.tsx` already uses `zodResolver(CreateTeamSchema)`, so we should just rely on `formState.errors` for field-level feedback. the server-side zod toast is redundant.
- `team-name.tsx` doesn't validate client-side — that's what's driving the need for this helper. let's move the schema check to the client (the input already has `maxLength`, plus the trim/empty guards) and treat any server `zodError` as an unexpected fallback with a generic toast. 
 
then we can delete `getTRPCValidationMessages` and `TrpcErrorWithZodDataSchema` entirely.

src/features/dashboard/settings/general/team-name.tsx
ben-fornefeld: pushed a few changes here:

- swapped the auto-shrink measurement effect for a fixed `text-2xl` heading with proper line-height. fixes the descender clipping at the previous 32px / 32px / `h-8` combo (g/p/y were getting cut off) and drops ~50 lines.
- the page was `flex-row` on all viewports, so the avatar ate half the width on iphone and the editing column got ~180px. stacked it on `< sm`.
- same idea for the InfoRow label/value — stacked on mobile, added `[overflow-wrap:anywhere]` so long emails wrap cleanly.
- bound the name label to the input via `htmlFor` / `id` for a11y.

let me know if you'd rather keep the original spec — i think dropping the auto-shrink in
favor of a smaller fixed heading is the cleaner trade since it removes the measurement
loop and fixes the descender bug in one go.

sarimrmalik: awesome, this is my preference too. made a small improvement, the "Add photo" button was fitting the full width on mobile, fixed

src/lib/utils/trpc-errors.ts
sarimrmalik: good point, fixed. for `team-name`, I also refactored to `zodResolver` for consistency.

ben-fornefeld (approved):
lgtm now!

PR #298, opened 2026-04-16
refactor: route team bootstrap through dashboard admin API

ben-fornefeld:
## Summary
- route dashboard team resolution through the dashboard API and pass the authenticated user id into `resolveUserTeam`
- switch user bootstrapping to the new `/admin/users/{userId}/bootstrap` contract and update generated API types
- split bootstrap logic into a dedicated admin users repository and keep user team reads scoped to the access token-backed teams repository

### DIFF
--- a/src/app/api/auth/callback/route.ts
+++ b/src/app/api/auth/callback/route.ts
@@ -1,6 +1,6 @@
 import { redirect } from 'next/navigation'
 import { AUTH_URLS, PROTECTED_URLS } from '@/configs/urls'
-import { createUserTeamsRepository } from '@/core/modules/teams/user-teams-repository.server'
+import { createAdminUsersRepository } from '@/core/modules/users/admin-repository.server'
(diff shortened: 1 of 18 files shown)


dobrac (approved):
LGTM! one question left

src/app/api/auth/callback/route.ts
```diff
@@ -49,10 +51,37 @@ export async function GET(request: Request) {
...
+
+      if (!data.session?.access_token || !userId) {
+        throw encodedRedirect(
+          'error',
+          AUTH_URLS.SIGN_IN,
+          'Missing session after auth callback'
+        )
+      }
+
+      if (ENABLE_USER_BOOTSTRAP) {
```
dobrac, line 64: why do we need this under a flag?

ben-fornefeld: only temporary for the migration period

PR #300, opened 2026-04-21
Feat: Sandbox details events table

sarimrmalik:
[Ticket](https://linear.app/e2b/issue/ENG-3828/sandbox-details-events-table)

## Summary
- Add a new "Events" tab to the sandbox details page that lists lifecycle events for the sandbox in a sortable table.
- Introduce `SandboxEventsPage`, `SandboxEventsTable`, and supporting cells (timestamp with centisecond precision + inline copy, event-type badge with per-type icon, ID badge with copy) built on the shared `Table` primitives for consistency with the rest of the dashboard.
- Add an `EventTypeFilter` so users can narrow events by lifecycle type, plus a clear empty state when no events match.
- Wire up routing in `urls.ts` and the dashboard layout tabs so the new page is reachable from the sandbox details view.

## Test plan
- Manually verified the events tab on the sandbox details page: sorting by timestamp, filtering by event type, copy buttons for timestamp and event ID, and the empty state when no events are returned.
- Confirmed the page renders correctly against the live sandbox events API on the Vercel preview deployment.

## Images
image


src/features/dashboard/settings/keys/id-badge.tsx
```diff
@@ -0,0 +1,43 @@
...
+const getIdBadgeLabel = (id: string): string => {
+  if (id.length <= 8) return id.toUpperCase()
+  return `${id.slice(0, 4)}...${id.slice(-4)}`.toUpperCase()
+}
+
+interface IdBadgeProps {
+  id: string
+}
+
+export const IdBadge = ({ id }: IdBadgeProps) => {
```
ben-fornefeld: this isn't keys-specific anymore — sandbox events imports it. let's move it to `src/features/dashboard/shared/id-badge.tsx`. also: now that api-keys is merged, `api-keys-table-row.tsx:71-87` inlines basically the same badge code, and `getApiKeyIdBadgeLabel` in `api-keys-utils.ts` is byte-identical to `getIdBadgeLabel` here. after the move we can have api-keys consume the shared component — separate follow-up is fine.

src/features/dashboard/sandbox/events/event-type-filter.tsx
```diff
@@ -0,0 +1,58 @@
+import {
```
ben-fornefeld, line 3: missing `'use client'`. uses radix dropdown + an `onValueChange` callback, and `view.tsx` / `table.tsx` are both client. inconsistent.

src/features/dashboard/sandbox/events/event-type-filter.tsx
```diff
@@ -0,0 +1,58 @@
...
+      <DropdownMenuContent align="start">
+        <DropdownMenuRadioGroup
+          value={type ?? ALL_EVENT_TYPES_VALUE}
+          onValueChange={(value) => {
+            if (value === ALL_EVENT_TYPES_VALUE) {
+              onTypeChange(null)
+              return
+            }
+
+            onTypeChange(SandboxLifecycleEventTypeSchema.parse(value))
```
ben-fornefeld: `.parse()` throws on unknown values. `event-type-badge.tsx:6` uses `.safeParse()` for the same enum. let's pick one — `safeParse` here too, fall back to the All branch on failure.

src/features/dashboard/sandbox/events/table.tsx
```diff
@@ -0,0 +1,127 @@
...
+  TableBody,
+  TableCell,
+  TableEmptyState,
+  TableHead,
+  TableHeader,
+  TableRow,
+} from '@/ui/primitives/table'
+import { SandboxEventTypeBadge } from './event-type-badge'
+
+const EVENT_COLUMN_WIDTHS = {
```
ben-fornefeld: let's use tailwind classes on `<col>` instead of inline styles via this helper — `api-keys-table.tsx` does this in main. also avoids the implicit `+ 16` padding coupling between the constants and css.

src/features/dashboard/sandbox/events/table.tsx
```diff
@@ -0,0 +1,127 @@
...
+    <Table className="min-w-[360px] table-fixed">
+      <colgroup>
+        <col style={colStyle(EVENT_COLUMN_WIDTHS.timestamp)} />
+        <col style={colStyle(EVENT_COLUMN_WIDTHS.id)} />
+        <col />
+      </colgroup>
+      <TableHeader className="bg-bg sticky top-0 z-10">
+        <TableRow>
+          <TableHead className="px-0 h-min text-fg" data-state="selected">
+            <button
```
ben-fornefeld: the sort toggle button has no hover/focus state and no padding — the tap target is text-sized. let's use `<Button variant="quaternary" size="sm">` or add proper styling.

src/features/dashboard/sandbox/events/table.tsx
```diff
@@ -0,0 +1,127 @@
...
+    </Table>
+  )
+}
+
+const EventsEmptyState = () => {
+  return (
+    <TableEmptyState colSpan={3}>
+      <div className="flex items-center gap-2">
+        <HistoryIcon className="size-5" />
+        <p className="prose-body-highlight text-fg-tertiary">No events found</p>
```
ben-fornefeld: after the api-keys merge, `TableEmptyState` auto-wraps children in `prose-body-highlight flex items-center justify-center gap-2`. the inner `<div>` and the `<p>` are both redundant — `<TableEmptyState colSpan={3}><HistoryIcon className="size-5" /> No events found</TableEmptyState>` is enough.

src/features/dashboard/sandbox/events/use-sandbox-event-filters.ts
```diff
@@ -0,0 +1,62 @@
...
+
+// Re-adds the "sandbox.lifecycle." prefix from a URL value, e.g. "created" -> "sandbox.lifecycle.created"
+const fromUrlValue = (
+  value: string | null
+): SandboxLifecycleEventType | null => {
+  if (!value) return null
+  const parsed = SandboxLifecycleEventTypeSchema.safeParse(
+    `sandbox.lifecycle.${value}`
+  )
+  return parsed.success ? parsed.data : null
```
ben-fornefeld: this silently swallows invalid `?type=` values — we return `null` but the bad value stays in the url. let's strip it via `setFilters({ type: null })` on parse failure so the url stays clean.

src/features/dashboard/sandbox/events/view.tsx
```diff
@@ -0,0 +1,42 @@
...
+  const { order, orderAsc, setOrder, setType, type } = useSandboxEventFilters()
+
+  const events = useMemo(() => {
+    const lifecycleEvents = sandboxLifecycle?.events ?? []
+    const filteredEvents = type
+      ? lifecycleEvents.filter((event) => event.type === type)
+      : lifecycleEvents
+    const orderedEvents = orderAsc
+      ? filteredEvents
+      : [...filteredEvents].reverse()
```
ben-fornefeld, line 28: this assumes `lifecycleEvents` is in asc order from the source — implicit contract. let's either sort explicitly with `.sort((a, b) => a.timestamp - b.timestamp)` or add a one-line comment about the source order.

src/core/modules/sandboxes/lifecycle-event-types.ts
```diff
@@ -0,0 +1,13 @@
+import { z } from 'zod'
+
+const SandboxLifecycleEventTypeSchema = z.enum([
```
ben-fornefeld, line 5: this duplicates `WEBHOOK_EVENTS` in `src/features/dashboard/settings/webhooks/constants.ts` — same 5 strings, different order. let's have webhooks consume `SandboxLifecycleEventTypeSchema.options` and drop `WEBHOOK_EVENTS` (the `WebhookEvent` type is unused too). bonus: the webhooks checkbox ui could use `SANDBOX_EVENT_TYPE_MAP` for friendlier labels — separate follow-up.

ben-fornefeld (changes requested):
second pass after the rebases. most of the original items are still open, plus one new consolidation: this pr's lifecycle enum duplicates `WEBHOOK_EVENTS` in webhooks settings. the api-keys merge also unlocks moving `IdBadge` to shared. see inline.

src/features/dashboard/sandbox/events/table.tsx
sarimrmalik: good catch, fixed

src/features/dashboard/sandbox/events/table.tsx
sarimrmalik: wanted to keep it consistent with the Logs view, so ended up using Button with quaternary variant to use the focus state and cursor-pointer and kept the look same as before

image

src/features/dashboard/sandbox/events/table.tsx
sarimrmalik: nice, was waiting for api-keys to merge to fix this, fixed

src/features/dashboard/sandbox/events/use-sandbox-event-filters.ts
sarimrmalik: nice, fixed

src/core/modules/sandboxes/lifecycle-event-types.ts
sarimrmalik: good catch, fixed

ben-fornefeld (changes requested):
a few general issues:
- we should have virtualization by default. check out the sandbox/template logs components and re-use where possible
- i cannot scroll the table right now if there is elements overflowing at the bottom. virtualization should address this

ben-fornefeld (changes requested):
can you check and address these concerns: 
- for the timestamp column width, it be determined based on the formatted string of now(). currently it appears to be truncated which is not ideal. it can change depending on which month / date is currently formatted.
- we should include the event data as a column. for visualization, we can show it the same way as metadata in the sandboxes list table.
- the filtering should be multi-checkboxable, meaning i can inspect multiple filtered event types not only select one event type or all

sarimrmalik:
@ben-fornefeld thanks for that, I just increased the timestamp width by 10px and then verified that for each month, the timestamp doesn't truncate.

Added event data + made event type filter multi-select.

image

ben-fornefeld (approved):

PR #305, opened 2026-04-28
Fix: Add recovery dialogs for missing payment method and verification required

sarimrmalik:
[Ticket](https://linear.app/e2b/issue/ENG-3846/handle-team-blockage-payment-method-missing-via-dashboard-dialog)

## Summary
- Show an in-app blocking dialog when the active team is blocked for a missing payment method.
- Add a billing tRPC mutation for creating payment method setup sessions and render the Stripe Payment Element in the dialog.
- Poll team state after successful setup and close the dialog once the missing-payment-method block clears.

## Test plan
- Manually verified the diff flow against the billing-server payment methods session contract.

## Images
- When team is blocked, the dialog automatically opens
image
- Clicking on banner also opens dialog
image
- Dialog when reason is `verification required`
image


src/features/dashboard/sidebar/missing-payment-method-dialog.tsx
```diff
@@ -0,0 +1,443 @@
+'use client'
```
ben-fornefeld: this dialog and `verification-required-dialog.tsx` are ~70-80% structurally identical, and `LoadingState` is also triplicated with `concurrent-sandboxes-addon-dialog.tsx`. lets extract a shared `team-blocked-recovery/` module with `useTeamUnblockPolling`, `LoadingState`, the `Elements` wrapper, and the stripe url-redirect handler. each dialog should then only specify its endpoint, copy, and confirm method (`confirmSetup` vs `confirmPayment`).

ben-fornefeld (changes requested):
context-gathering pass on the recovery dialogs. headline asks: extract the shared logic so the two dialogs become small adapters, and move them out of `sidebar/` since they're not sidebar-specific. inline notes for the per-file items.

ben-fornefeld (commented):
(probe — github 422 debug, ignore)

ben-fornefeld (commented):
(probe — github 422 debug, ignore)

ben-fornefeld (commented):
(probe — github 422 debug, ignore)

ben-fornefeld (commented):
(probe — github 422 debug, ignore)

ben-fornefeld (commented):
(probe — github 422 debug, ignore)

ben-fornefeld (commented):
(probe — github 422 debug, ignore)

src/features/dashboard/sidebar/verification-required-dialog.tsx
```diff
@@ -0,0 +1,476 @@
+'use client'
+
+import {
+  Elements,
+  PaymentElement,
```
ben-fornefeld: both recovery dialogs live in `sidebar/` but they're rendered by `layouts/team-blocked-indicator.tsx` and have nothing to do with the sidebar. let's move them next to the addon dialog in `billing/`, or a new `team-blocked/` folder if we want to keep `billing/` focused on subscription stuff.

src/features/dashboard/sidebar/verification-required-dialog.tsx
```diff
@@ -0,0 +1,476 @@
...
+    router,
+    onOpenChange,
+  ])
+
+  return (
+    <>
+      <DialogHeader>
+        <DialogTitle>Verify team</DialogTitle>
+        <DialogDescription>
+          This team is blocked because verification is required. Make a $5
```
ben-fornefeld: `$5` is hardcoded here and again at line 433. if the server amount changes, two strings drift. let's either return the amount in `VerificationPaymentResponse` or hoist to a single constant.

src/features/dashboard/sidebar/missing-payment-method-dialog.tsx
```diff
@@ -0,0 +1,443 @@
...
+    paymentMethodsSessionMutation.reset()
+  }, [open, paymentMethodsSessionMutation.reset])
+
+  useEffect(() => {
+    if (!open) return
+
+    const setupIntentClientSecret = setupIntentParams.setup_intent_client_secret
+
+    if (hasHandledSetupIntent.current) return
+    hasHandledSetupIntent.current = true
```
ben-fornefeld: flipping the ref to `true` before the async work runs is fragile — if `checkSetupIntent` fails before resolving, the ref is already set and the same mount can't retry. let's flip it after the async resolves, or move to a state-machine style with explicit success/error states.

src/features/dashboard/layouts/team-blocked-indicator.tsx
```diff
@@ -44,25 +51,63 @@ function useBlockedMessage(slug: string, blockedReason: string | null) {
...
+  const [openDialog, setOpenDialog] = useState<BlockedReasonDialog>(null)
 
   const message = useBlockedMessage(team.slug, team.blockedReason)
+  const reason = team.blockedReason?.toLowerCase() ?? ''
+  const blockedReasonDialog: BlockedReasonDialog =
+    Object.values(BLOCKED_REASONS).find((blockedReason) =>
+      reason.includes(blockedReason)
+    ) ?? null
+
+  useEffect(() => {
```
ben-fornefeld, line 90: this re-opens the dialog on every mount — including after client-side navigation. if the user dismissed it, that gets ignored. let's persist a `sessionStorage` flag per blocked reason and skip the auto-open when it's set. clicking the banner cta should still open the dialog regardless — dismissable per session, but always reopenable via the banner.

ben-fornefeld (commented):
(continuation of review above — part 1)

src/features/dashboard/sidebar/missing-payment-method-dialog.tsx
```diff
@@ -0,0 +1,443 @@
...
+  DialogHeader,
+  DialogTitle,
+} from '@/ui/primitives/dialog'
+import { ArrowRightIcon, CardIcon } from '@/ui/primitives/icons'
+import { Loader } from '@/ui/primitives/loader'
+import { stripePromise, usePaymentElementAppearance } from '../billing/hooks'
+import { useDashboard } from '../context'
+
+const TEAM_UNBLOCK_POLL_ATTEMPTS = 15
+const TEAM_UNBLOCK_POLL_INTERVAL_MS = 2000
```
ben-fornefeld: `2000ms` between polls feels long for an in-flight unblock check — let's drop it (e.g. `750-1000ms`) so user feedback tightens up. total cap can stay roughly the same by bumping `TEAM_UNBLOCK_POLL_ATTEMPTS`. also: this constant is duplicated in `verification-required-dialog.tsx:33` — should live in the shared module from the M1 comment.

src/features/dashboard/sidebar/missing-payment-method-dialog.tsx
```diff
@@ -0,0 +1,443 @@
...
+      })
+      setPaymentMethodsSession(session)
+    } catch {
+      // The mutation onError handler owns the user-facing toast and close.
+    } finally {
+      setIsLoadingPaymentMethodsSession(false)
+    }
+  }, [paymentMethodsSessionMutation.mutateAsync, team.slug])
+
+  useEffect(() => {
```
ben-fornefeld: this effect is mostly redundant — radix unmounts `DialogContent` on close, so the next open remounts with fresh state anyway. `mutation.reset()` also doesn't abort in-flight fetches. let's drop the effect, or wire an `AbortController` if we genuinely want to cancel pending requests.

src/features/dashboard/layouts/team-blocked-indicator.tsx
```diff
@@ -44,25 +51,63 @@ function useBlockedMessage(slug: string, blockedReason: string | null) {
...
+          {message.text}
+          {message.cta && (
+            <>
+              {' '}
+              {message.href ? (
+                <Link href={message.href} className="underline">
+                  {message.cta}
+                </Link>
+              ) : (
+                <button
```
ben-fornefeld: let's use the `<Button>` primitive (e.g. `variant="quaternary"`) instead of a raw `<button>` with manual `cursor-pointer underline`. matches the rest of the codebase.

src/features/dashboard/sidebar/missing-payment-method-dialog.tsx
```diff
@@ -0,0 +1,443 @@
...
+      toast(
+        defaultErrorToast(
+          error.message ?? 'Failed to save payment method. Please try again.'
+        )
+      )
+      setIsSaving(false)
+      return
+    }
+
+    toast({
```
ben-fornefeld: the addon dialog tracks payment events via posthog — recovery flows don't. recovery is a high-value funnel; let's add `posthog.capture('payment_method_added')` here and the verification equivalent so we can see drop-off.

ben-fornefeld (commented):
(continuation of review above — part 2)

src/features/dashboard/sidebar/verification-required-dialog.tsx
sarimrmalik: good catch, the team blocked indicator was initially in the sidebar, that's why i was using the sidebar directory. fixed.

src/features/dashboard/sidebar/verification-required-dialog.tsx
sarimrmalik: the better solution is to return the amount in `VerificationPaymentResponse` so down the road we can vary this entirely from the billing server, implemented

src/features/dashboard/layouts/team-blocked-indicator.tsx
sarimrmalik: nice, i created a reusable hook for this `useSessionStorage`

src/features/dashboard/sidebar/missing-payment-method-dialog.tsx
sarimrmalik: fair, fixed

src/features/dashboard/layouts/team-blocked-indicator.tsx
sarimrmalik: i did it but to use `quaternary` here you almost have to override everything which is not the best, perhaps there is an opportunity for a `ghost` button variant that doesn't come with any css, just `cursor-pointer `etc. for later use.

src/features/dashboard/sidebar/missing-payment-method-dialog.tsx
sarimrmalik: nice, added

src/features/dashboard/sidebar/missing-payment-method-dialog.tsx
sarimrmalik: using a state-machine style setup

src/lib/hooks/use-session-storage.ts
ben-fornefeld, line 1: please use https://usehooks-ts.com/react-hook/use-session-storage

src/features/dashboard/layouts/team-blocked-indicator.tsx
```diff
@@ -44,25 +53,92 @@ function useBlockedMessage(slug: string, blockedReason: string | null) {
...
+    setOpenDialog(blockedReasonDialog)
+  }, [blockedReasonDialog, dismissedStorageKey, dismissedStorage])
+
+  const handleDialogAction = () => {
+    setOpenDialog(blockedReasonDialog)
+  }
+
+  const handleDialogOpenChange = (open: boolean, dialog: TeamBlockedReason) => {
+    if (!open) {
+      dismissedStorage.setValue('true')
```
ben-fornefeld: this writes `'true'` on every close, including success-triggered ones. the dialogs work around it by calling `removeValue()` *after* `onOpenChange(false)` (e.g. `missing-payment-method-dialog.tsx:140-141`), so the success cleanup undoes the dismissal. but reorder those two lines for any reason and the success path silently flips into 'permanently dismissed for the session'.

let's only mark dismissed when the close is user-initiated. e.g. pass an explicit `reason: 'user' \| 'success'` to the close handler, or have the dialog itself call `setValue('true')` on user dismissals only.

ben-fornefeld (changes requested):
one more thing — the dismissed-flag ordering is fragile. inline comment.

src/lib/hooks/use-session-storage.ts
sarimrmalik: done

ben-fornefeld (approved):
lgtm

PR #329, opened 2026-05-14
fix: prevent sensitive data leaking into logs

drankou:
Closes #259 . Supersedes #260 with a broader fix + tests.

## Changes
- `src/app/api/auth/callback/route.ts` — error branch logs `hasCode: !!code `instead of the raw auth code (matches the info log above).
- `src/core/shared/clients/logger/redaction.ts` — add `signatureSecret` to redaction paths. (access_token is already covered.)
- `src/core/server/actions/{client,utils}.ts` —  avoid logging raw `clientInput`. A new `summarizeClientInput` helper inlines a small set of allowlisted scalar keys and replaces all other fields with shape hints like `string(64)`, `array(3)`, `object`.

### DIFF
--- a/src/core/server/actions/client.ts
+++ b/src/core/server/actions/client.ts
@@ -28,7 +28,7 @@ import type {
 import {
   ActionError,
   flattenClientInputValue,
-  summarizeClientInput,
+  sanitizeClientInput,
 } from './utils'
 
 type SupabaseServerClient = Awaited<ReturnType<typeof createClient>>
@@ -139,11 +139,7 @@ export const actionClient = createSafeActionClient({
   const baseLogPayload = {
     server_function_type: type,
     server_function_name: name,
-    // Do NOT log raw clientInput — it's an arbitrary user-supplied payload
-    // and relying on pino redaction as a blocklist is fragile (any new action
-    // with a non-standard sensitive field name would leak by default).
-    // summarizeClientInput keeps an allowlisted, debug-safe shape view.
-    server_function_input_summary: summarizeClientInput(clientInput),
+    server_function_input_summary: sanitizeClientInput(clientInput),
     server_function_duration_ms: duration.toFixed(3),
     request_url: requestObservabilityContext.request_url,
     request_path: requestObservabilityContext.request_path,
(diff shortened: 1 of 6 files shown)


src/core/server/actions/client.ts
```diff
@@ -135,7 +139,11 @@ export const actionClient = createSafeActionClient({
   const baseLogPayload = {
     server_function_type: type,
     server_function_name: name,
-    server_function_input: clientInput,
+    // Do NOT log raw clientInput — it's an arbitrary user-supplied payload
+    // and relying on pino redaction as a blocklist is fragile (any new action
+    // with a non-standard sensitive field name would leak by default).
+    // summarizeClientInput keeps an allowlisted, debug-safe shape view.
```
ben-fornefeld: i don't think we need the comment

src/core/server/actions/utils.ts
```diff
@@ -44,3 +44,73 @@ export const flattenClientInputValue = (
...
+ *
+ * - Allowlisted scalar keys are inlined as-is.
+ * - Everything else is replaced with `_<key>: <type-hint>` so shape and
+ *   length are visible without leaking values.
+ *
+ * This is the inverse of the previous blocklist-via-pino-redaction approach:
+ * new sensitive fields are safe by default and only become loggable when
+ * explicitly added to {@link SAFE_INPUT_KEYS}.
+ */
+export function summarizeClientInput(input: unknown): Record<string, unknown> {
```
ben-fornefeld:
```suggestion
export function sanitizeClientInput(input: unknown): Record<string, unknown> {
```

src/core/server/actions/utils.ts
```diff
@@ -44,3 +44,73 @@ export const flattenClientInputValue = (
...
 }
+
+/**
+ * Keys that are safe to log in cleartext from action inputs — IDs, slugs,
+ * enums, and small flags. Any key not on this list is replaced with a type
+ * hint (e.g. `string(64)`, `array(3)`, `object`) so we never leak raw values.
+ *
+ * Keep this list tight: prefer adding scoped per-action logging in the action
+ * body over expanding this allowlist.
+ */
```
ben-fornefeld: we don't need the comment

ben-fornefeld (approved):
legit

PR #332, opened 2026-05-15
fix: auth redirect query encoding

drankou:
## Summary

Fix auth redirect message handling so query parameters are encoded once via `URLSearchParams` instead of manually encoding/decoding values across server and client code.

This updates the email callback redirect to use the shared `encodedRedirect()` helper for `message` redirects, preventing crafted `message` values from injecting sibling query parameters.

## Changes

- Extend `encodedRedirect()` to support `message` redirects.
- Let `URLSearchParams` handle redirect query encoding directly.
- Replace raw email-callback `redirect(...?message=${message}...)` calls with `encodedRedirect('message', ...)`.
- Remove unnecessary `decodeURIComponent(...)` calls from auth/account message rendering.
- Add regression coverage for query parameter injection in `/api/auth/email-callback`.

Closes #255

### DIFF
--- a/src/app/(auth)/auth/cli/page.tsx
+++ b/src/app/(auth)/auth/cli/page.tsx
@@ -177,7 +177,7 @@ export default async function CLIAuthPage({
       <div className="text-fg-tertiary mt-12 leading-8">
         <Suspense fallback={<div>Loading...</div>}>
           {error ? (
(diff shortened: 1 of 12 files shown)


ben-fornefeld (approved):
lgtm,

small nit in `src/core/server/actions/user-actions.ts`: can we build `emailRedirectTo` with `URL` + `searchParams.set()` too?

PR #352, opened 2026-06-02
feat: templates list pagination

drankou:
## Summary

Replaces the full-list `getTemplates` query with a cursor-paginated endpoint and rewires the templates list UI around `useSuspenseInfiniteQuery` + a Load More button. Sorting, filtering, and search are now executed server-side, so the page no longer needs to download every template up front and the table is a pure renderer over the rows returned by the paginated query.

### Backend (tRPC + repository)
- New `listTeamTemplates` repo method accepting `cursor`, `limit`, `cpuCount`, `memoryMB`, `public`, `search`, and a single `sort` token; returns `{ data, nextCursor }`.
- `templates.getTemplates` router input now validates the same params (limit ≤ 100, default 50; `sort` defaults to `updated_at_desc`).
- Full-list `getTeamTemplates` is kept untouched — still used by the terminal page.

### Sort token
The active react-table column + direction is mapped to one of the server enum values:

| Column      | Asc                | Desc                  |
|-------------|--------------------|-----------------------|
| `name`      | `name_asc`         | `name_desc`           |
| `cpuCount`  | `cpu_count_asc`    | `cpu_count_desc`      |
| `memoryMB`  | `memory_mb_asc`    | `memory_mb_desc`      |
| `createdAt` | `created_at_asc`   | `created_at_desc`     |
| `updatedAt` | `updated_at_asc`   | `updated_at_desc`     |

### DIFF
--- a/src/core/modules/templates/models.ts
+++ b/src/core/modules/templates/models.ts
@@ -23,3 +23,30 @@ export type DefaultTemplate = Template & {
   isDefault: true
   defaultDescription?: string
 }
+
+export type TemplatesSort =
+  | 'name_asc'
+  | 'name_desc'
+  | 'cpu_count_asc'
+  | 'cpu_count_desc'
+  | 'memory_mb_asc'
+  | 'memory_mb_desc'
+  | 'created_at_asc'
+  | 'created_at_desc'
+  | 'updated_at_asc'
+  | 'updated_at_desc'
+
+export interface ListTeamTemplatesOptions {
+  cursor?: string
+  limit?: number
+  cpuCount?: number
+  memoryMB?: number
+  public?: boolean
+  search?: string
+  sort?: TemplatesSort
+}
+
+export interface ListTeamTemplatesResult {
+  data: Array<Template | DefaultTemplate>
+  nextCursor: string | null
+}
--- a/src/core/modules/templates/models.ts
+++ b/src/core/modules/templates/models.ts
@@ -1,3 +1,7 @@
+import type {
+  components as DashboardComponents,
+  paths as DashboardPaths,
+} from '@/contracts/dashboard-api'
 import type { components as InfraComponents } from '@/contracts/infra-api'
 
 export type Template = Pick<
@@ -24,27 +28,11 @@ export type DefaultTemplate = Template & {
   defaultDescription?: string
 }
 
-export type TemplatesSort =
-  | 'name_asc'
-  | 'name_desc'
-  | 'cpu_count_asc'
-  | 'cpu_count_desc'
-  | 'memory_mb_asc'
-  | 'memory_mb_desc'
-  | 'created_at_asc'
-  | 'created_at_desc'
-  | 'updated_at_asc'
-  | 'updated_at_desc'
+export type TemplatesSort = DashboardComponents['parameters']['templates_sort']
 
-export interface ListTeamTemplatesOptions {
-  cursor?: string
-  limit?: number
-  cpuCount?: number
-  memoryMB?: number
-  public?: boolean
-  search?: string
-  sort?: TemplatesSort
-}
+export type ListTeamTemplatesOptions = NonNullable<
+  DashboardPaths['/templates']['get']['parameters']['query']
+>
 
 export interface ListTeamTemplatesResult {
   data: Array<Template | DefaultTemplate>
--- a/src/core/modules/templates/repository.server.ts
+++ b/src/core/modules/templates/repository.server.ts
@@ -102,15 +102,7 @@ export function createTemplatesRepository(
 
       const res = await deps.apiClient.GET('/templates', {
         params: {
-          query: {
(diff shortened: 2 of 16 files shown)


src/features/dashboard/templates/list/table.tsx
```diff
@@ -30,47 +36,81 @@ import { useTemplateTableStore } from './stores/table-store'
...
+const PAGE_SIZE = 50
+
+const COLUMN_TO_SORT_BASE: Record<string, string> = {
+  name: 'name',
+  cpuCount: 'cpu_count',
+  memoryMB: 'memory_mb',
+  createdAt: 'created_at',
+  updatedAt: 'updated_at',
+}
+
```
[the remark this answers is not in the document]
drankou: should be fine for now, most sorted are cpu/memory (90% of posthog events)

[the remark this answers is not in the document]
drankou: disabled sort by id

src/core/modules/templates/models.ts
```diff
@@ -23,3 +23,30 @@ export type DefaultTemplate = Template & {
...
+  memoryMB?: number
+  public?: boolean
+  search?: string
+  sort?: TemplatesSort
+}
+
+export interface ListTeamTemplatesResult {
+  data: Array<Template | DefaultTemplate>
+  nextCursor: string | null
+}
```
ben-fornefeld: do we need to re-define those ourselves or can we infra from oapi generated types

src/features/dashboard/templates/list/constants.ts
```diff
@@ -0,0 +1,16 @@
+import type { SortingState } from '@tanstack/react-table'
+import type { TemplatesSort } from '@/core/modules/templates/models'
+
+export const TEMPLATES_PAGE_SIZE = 50
```
huv1k, line 4: Should we have some default page size for everything instead of having it specific to templates?

src/core/modules/templates/repository.server.ts
```diff
@@ -87,6 +95,76 @@ export function createTemplatesRepository(
...
+        params: {
+          query: {
+            cursor: options.cursor,
+            limit: options.limit,
+            cpuCount: options.cpuCount,
+            memoryMB: options.memoryMB,
+            public: options.public,
+            search: options.search,
+            sort: options.sort,
+          },
```
huv1k: Is there any reason to have it explicit like this? 
```suggestion
          query: options,
```

src/core/modules/templates/repository.server.ts
```diff
@@ -87,6 +95,76 @@ export function createTemplatesRepository(
...
+            // Email resolution is deferred while the Supabase auth migration is
+            // in progress; the endpoint returns only the creator id for now.
+            createdBy: t.createdBy
+              ? { id: t.createdBy.id, email: t.createdBy.email ?? '' }
+              : null,
+            lastSpawnedAt: t.lastSpawnedAt ?? null,
+            spawnCount: t.spawnCount,
+            buildCount: t.buildCount,
+            envdVersion: t.envdVersion ?? '',
+          }
```
huv1k: Do we need to create another object?

src/core/modules/templates/repository.server.ts
```diff
@@ -87,6 +95,76 @@ export function createTemplatesRepository(
...
+        return err(
+          repoErrorFromHttp(
+            res.response.status,
+            res.error?.message ?? 'Failed to fetch templates',
+            res.error
+          )
+        )
+      }
+
+      const data = (res.data?.data ?? []).map(
```
huv1k: I think early return would be better then mapping over empty array

huv1k (commented):
Few nit picks, what do you think?

drankou:
@huv1k addressed all three:

- Collapsed the query to `query: options` — type-checks since `ListTeamTemplatesOptions` and the OpenAPI query are field-identical.
- Replaced `base` + spread with a single object literal using a conditional spread for the `DefaultTemplate` discriminator (one allocation per item instead of two on the default path).
- Added an explicit early return for the empty-data case so the mapping path only runs when there's something to map.

Skipping the page-size constant nit for now — happy to lift it to a shared constant in a follow-up if other paginated lists land that'd reuse it.

Pushed in fa088aba.

src/features/dashboard/templates/list/constants.ts
drankou: I would keep it as-is for now since page size could depend on row item size and API performance

src/core/modules/templates/repository.server.ts
drankou: addressed

ben-fornefeld (approved):
one nit, otherwise lgtm

src/core/modules/templates/models.ts
drankou: adressed

PR #353, opened 2026-06-04
refactor: server actions to trpc

sarimrmalik:


sarimrmalik:
Closing since not a priority right now while we refactor the auth logic

PR #381, opened 2026-06-10
refactor(sandbox): move sandbox state management server-side (ENG-4287)

huv1k:
Moves sandbox state management server-side so the user's **account-level access token is never exposed to browser JS** (ENG-4287), and adds the ability to **pause and resume a sandbox** from the dashboard. Previously the sandbox-details surfaces drove the E2B SDK client-side with `sandboxManagementAuth` (the account token); now the browser only ever receives **sandbox-scoped envd credentials**.

### Server-side state management
- **Filesystem inspector** builds an **envd-only** SDK client via a shared `createEnvdSandbox` helper from the `envdAccessToken` already returned by the `sandbox.details` query — no control-plane call, so normal inspect never resumes a paused sandbox or extends its TTL.
- **Terminal** (standalone `/dashboard/terminal` **and** the sandbox-details terminal tab) runs create/connect through a **`sandbox.openTerminal`** tRPC mutation; the client receives only scoped envd creds.
- **Explicit "Resume sandbox"** (inspector) runs its control-plane connect through a new **`sandbox.resume`** tRPC mutation server-side (short `SANDBOX_RESUME_TIMEOUT_MS` window), then the client rebuilds the envd-only client from the returned creds. Resuming/extending TTL only happens on this explicit user action — never implicitly.
- Tokenless (`secure: false`) sandboxes are supported end-to-end (`envdAccessToken` optional throughout).
- The now-unused `sandbox-management-auth` modules are deleted; `getFullInfo`/`sandbox.details` remain read-only.

### Pause / resume
- **Pause button** next to Kill in the sandbox header, shown **only while the sandbox is running**, with a confirmation popover.
- Pause runs server-side through a new **`sandbox.pause`** tRPC mutation using the SDK's `Sandbox.pause()` (consistent with `sandbox.resume`), so the account token stays off the client.
- **Resume** a paused sandbox from the inspect Filesystem/Terminal empty states; the empty state notes the resume window and that the sandbox can be paused again.
- **Prevents auto-resume on pause:** a `memory: true` pause can be auto-resumed by *any* envd traffic, so the live terminal PTY and filesystem watcher would immediately revive the sandbox. The Pause button now **optimistically flips state to `paused`**, tearing those connections down *before* the snapshot runs (rolled back on error); without the live traffic the snapshot completes and the sandbox stays paused.
- Shows the **"Sandbox Paused"** empty state for both the filesystem and terminal instead of stale cached data.

### Notes
- Transport choice: terminal/resume/pause mutations use tRPC (vanilla client injected into the helpers), consistent with the sandbox router's existing queries.
- `origin/main` merged in — reconciled main's resume UX and details terminal tab onto the no-token model, and adopted main's removal of the client-side attach-retry.

Tests: typecheck/lint clean on changed files.

**Not yet verified live:** on the preview deploy, (1) confirm via DevTools that envd requests carry only `X-Access-Token` — no account `Authorization: Bearer` / `X-Supabase-Token` (incl. an explicit Resume and a `secure: false` sandbox); (2) from the **Terminal** and **Filesystem** tabs, click **Pause** and confirm the sandbox shows the paused screen and **does not auto-resume**.

**Known follow-up (pre-existing, from main):** `sandbox.killTerminalPty` still does a control-plane `Sandbox.connect` (resume + extend TTL) just to kill a PTY on terminal close; it should become envd-only to avoid that side effect.

| Before | After |
| ------ | ----- |
| CleanShot 2026-06-25 at 19 44 57@2x  | CleanShot 2026-06-25 at 19 45 13@2x  |
| CleanShot 2026-06-25 at 19 45 55@2x | CleanShot 2026-06-25 at 19 46 30@2x |


huv1k:
Addressed both Codex P2 comments in 37592def: `envdAccessToken` is now optional end-to-end, so `secure: false` sandboxes (envd reachable without `X-Access-Token`) and sandboxes with a null `domain` connect for both the terminal and the filesystem inspector. Removed the over-strict guards and added a tokenless-sandbox unit test.

@codex review

ben-fornefeld (approved):
lgtm!

PR #382, opened 2026-06-10
fix(auth): keep Auth.js session handling away from non-auth requests

huv1k:
## Summary

Two related fixes preventing stray requests from touching auth state:

1. **Matcher: skip all `_next/*` paths.** Only `_next/static` and `_next/image` were excluded, so other `/_next/*` paths (e.g. stray `/_next/mintlify-assets/*` requests observed in traffic) ran the full middleware before 404ing. Safe because the app is App Router only (no `_next/data`), and server actions / RSC navigations use page paths. Root-level `mintlify-assets/` and `_mintlify/` exclusions are kept for the `next.config.ts` rewrites.

2. **Scope the Auth.js wrapper to dashboard + auth routes.** In Ory mode the wrapper ran on every matched path, re-setting the session cookie per request and deleting tokens it cannot decrypt — observed in production as `__Secure-authjs.session-token=; Max-Age=0` on a 404ing asset request, i.e. a stray request could log the user out. Only `/dashboard/*` and the auth pages consume the resolved session (`handleAuthGate` / `getOryAuthRouteRedirect`); everything else now goes through plain `proxyCore` with no session side effects.

## Testing

- New tests assert the wrapper runs for `/dashboard/*`, `/sign-in`, `/sign-up` and is skipped for `/`, `/blog/*`, stray `_next` paths
- `tests/unit/proxy-handlers.test.ts`, `tests/integration/proxy.test.ts`, `tests/integration/auth-ory-entrypoints.test.ts` — 20/20 pass
- Verified on the branch preview deployment: `/_next/mintlify-assets/...` returns a bare 404 with no `set-cookie`, while the old build on e2b-juliett.dev sets/deletes authjs cookies on the same path

### DIFF
--- a/src/proxy.ts
+++ b/src/proxy.ts
@@ -12,6 +12,8 @@ import {
   handleMiddlewareRedirect,
   handleMiddlewareRewrite,
   handleRouteRewritePassthrough,
+  isAuthRoute,
+  isDashboardRoute,
 } from './core/server/http/proxy'
 import { l, serializeErrorForLog } from './core/shared/clients/logger/logger'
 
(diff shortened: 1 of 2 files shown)


ben-fornefeld:
closing this in favor of https://github.com/e2b-dev/dashboard/pull/384

PR #398, opened 2026-06-11
fix(auth): clear session in a plain route handler so logout sticks

huv1k:
## What

Replace the `signOutAction()` server-action logout with a dedicated `GET /api/auth/sign-out` route handler that the logout button hard-navigates to. Removes the now-unused `signOutAction`, simplifies `completeOrySignOut` to skip the Ory RP-initiated logout round-trip, and keeps the PostHog identity reset on logout.

## Why

The dashboard session is a self-contained JWT cookie, so logout only takes effect if that cookie's `Set-Cookie` deletion reaches the browser. When sign-out runs inside an `auth()`-wrapped request, the Auth.js wrapper re-issues a refreshed session cookie at the end of the request, clobbering the deletion and leaving the user logged in. A plain, unwrapped route handler makes the deletion the final word. The hard navigation also keeps the "Logging out…" overlay up until the page unloads (no soft RSC redirect flash), and the Ory path now redirects straight to the post-logout page, relying on the existing Kratos + Hydra admin revocations. Integration tests exercise the new route.

| Before | After |
| ------ | ----- |
| CleanShot 2026-06-12 at 11 56 01 | CleanShot 2026-06-12 at 11 57 45  |

## Note

The root cause is environment-agnostic, so this fixes logout on **preview deployments** as well as production — clearing the cookie in an unwrapped route handler makes the deletion stick regardless of `VERCEL_ENV`.


src/core/server/auth/ory/signout-flow.ts
```diff
@@ -53,6 +49,8 @@ export async function completeOrySignOut(origin = BASE_URL): Promise<string> {
     identityId ? revokeKratosSessionsForIdentity(identityId) : null,
   ])
 
-  const logoutUrl = idToken ? buildOryLogoutUrl({ idToken, origin }) : null
-  return (logoutUrl ?? postLogoutUrl).toString()
+  // The admin revocations above already terminate the Kratos + Hydra sessions
+  // server-side, so we skip the Ory RP-initiated logout round-trip and redirect
+  // straight to the post-logout page — one redirect, no cross-origin hop.
```
ben-fornefeld: remove this comment please

ben-fornefeld (approved):

src/core/server/auth/ory/signout-flow.ts
ben-fornefeld:
```suggestion
```

PR #411, opened 2026-06-15
feat: make templates list build-agnostic, move resources to builds

drankou:
Templates list no longer shows build-derived columns; builds page now carries the per-build resources and surfaces failure reasons inline.

- templates list: drop CPU, Memory, ENVD columns
- builds: add CPU, Memory, Storage, ENVD columns (sorting disabled)
- builds: drop the Reason column; show the error in a tooltip on the Failed badge


huv1k (approved):
I would add tooltip, to explain what is ENVD Ver., otherwise it looks good

Image

PR #414, opened 2026-06-15
feat(auth): restore custom Ory Elements login/registration UI

huv1k:
Restores the custom Ory Elements login/registration UI originally landed in #380 and reverted in #399.

## What this brings back
- Custom Ory Elements login & registration pages (`src/app/login/**`, `src/app/registration/**`)
- SSO provider buttons with inline GitHub / Google logos and an "or" divider
- Ory proxy/config wiring (`src/configs/ory.ts`, `src/proxy.ts`)

## Beyond the original #380
Two follow-up commits made after the original PR are included:
- `16c3e0b9` — gate the custom Ory UI behind `NEXT_PUBLIC_ORY_CUSTOM_UI`
- `4c702d54` — forward Ory SDK paths to Kratos before the auth-session bypass

## Staging/preview redirect fix (separate commit)
- `c9475746` — patch `@ory/nextjs` to honor the request host on Vercel preview.

  `@ory/nextjs@1.0.0-rc.1`'s `guessPotentiallyProxiedOrySdkUrl()` returns `https://$VERCEL_URL` whenever `VERCEL_ENV` is set and the env is not production, so on staging (`VERCEL_ENV=preview`) the login flow's base URL became the per-deployment generated URL (`web-*.e2b-preview.dev`) instead of the canonical `e2b-staging.dev`. That bounced legs 3–6 of the auth chain (`/self-service` redirect, flow restarts, form-action rewrites, post-login landing) off the visited origin.

  The flow getters already pass the correct request-derived origin as `knownProxiedUrl` (`getPublicUrl()`), but the `VERCEL_URL` branch was checked first. The patch reorders `guessPotentiallyProxiedOrySdkUrl` to honor an explicit `knownProxiedUrl` before the Vercel env guesses, across all four compiled copies (app/pages × cjs/mjs).

  Applied as a temporary local patch via `bun patch` (version-pinned to `1.0.0-rc.1`); an upstream change to make the library prefer the request host will follow.

## Notes
- Updated to current `main`; `bun.lock` regenerated (only lockfile conflict on merge).
- Opened as draft for re-review before re-landing.
- `tests/integration/proxy.test.ts` passing locally (11/11).


huv1k:
Closing — superseded by #428 (custom Ory Elements flow pages, merged) and the broader auth direction is now the Kratos-only PR #427.

PR #426, opened 2026-06-17
Fix/filesystem querykey

matthewlouisbrockman:
reload files on tab change


ben-fornefeld (approved):
it's a bit of a smell that we should refactor the state management for this route, but we'll get to it

PR #428, opened 2026-06-17
feat(auth): env-guarded custom Ory Elements flow pages (all four flows)

huv1k:
Renders the **login, registration, recovery, and verification** Ory self-service flows with a custom `@ory/elements-react` UI served same-origin, so the whole flow (including preview deployments) stays on the visiting host. Gated by `NEXT_PUBLIC_ORY_CUSTOM_UI`: enabled on Preview + Staging, while on Production (unset) the pages redirect to the existing `/sign-in`, `/sign-up`, `/forgot-password` — and the gated same-origin Ory SDK proxy (`/self-service`, `/sessions/whoami`, …) is inert there, so the Auth.js + Hydra path is untouched. Extracts the UI from #414, extends it to all four flows (like #427), and swaps #414's PostHog gate for the plain env var. Adds `@ory/elements-react` + `@ory/nextjs`, `src/configs/ory.ts`, the component overrides (including a single-input `CodeInput` and flow-aware card header/footer), and the four flow pages.

**Follow-ups (not in this PR):** the `@ory/nextjs` preview-host patch (so the SDK URL honors the request host on Vercel previews rather than `VERCEL_URL`) is intentionally left out and will land separately. Ory/Hydra's server-side `*_ui_url` must also point at the dashboard origin for staging/preview to route here (terraform `enable_custom_auth_ui`); set `NEXT_PUBLIC_ORY_CUSTOM_UI=true` + `NEXT_PUBLIC_ORY_SDK_URL` in those envs.


src/app/registration/page.tsx
```diff
@@ -0,0 +1,31 @@
...
+  if (!flow) {
+    return null
+  }
+
+  const requestHeaders = await headers()
+  const host = requestHeaders.get('host')
+  const proto = requestHeaders.get('x-forwarded-proto') ?? 'http'
+  const config = host
+    ? { ...oryConfig, sdk: { ...oryConfig.sdk, url: `${proto}://${host}` } }
+    : oryConfig
```
drankou: since it mirrors  @ory/nextjs `getPublicUrl()`, we can also extract it into helper function

src/app/login/components/custom-code-input.tsx
```diff
@@ -0,0 +1,17 @@
...
+// Render OTP/recovery codes as a single dashboard Input: the default segmented
+// box group relies on theme/styles.css, which we don't import.
+export function OryCodeInput({ inputProps }: OryNodeInputProps) {
+  return (
+    <Input
+      {...inputProps}
+      inputMode="numeric"
+      autoComplete="one-time-code"
+      placeholder="123456"
+    />
```
drankou: we could use [InputOtp](<https://ui.shadcn.com/docs/components/radix/input-otp>) for this right?

src/app/login/components/custom-sso-button.tsx
```diff
@@ -0,0 +1,73 @@
...
+    provider,
+    node.meta?.label?.text ?? '',
+  ]
+    .join(' ')
+    .toLowerCase()
+
+  if (haystack.includes('github')) return PROVIDERS.github
+  if (haystack.includes('google')) return PROVIDERS.google
+  return undefined
+}
```
drankou: ory doesn't provide helpers for such resolution?

huv1k: It doesn't sadly 🙈

src/app/login/components/custom-code-input.tsx
huv1k: Yes, refactored

drankou (approved):

PR #429, opened 2026-06-17
fix(auth): patch @ory/nextjs to honor request host on preview deploys

huv1k:
## What

Adds the `@ory/nextjs` preview-host patch that #428 intentionally deferred. Reorders `guessPotentiallyProxiedOrySdkUrl` to honor `knownProxiedUrl` (the per-request host, already supplied via the package's `getPublicUrl()`) **before** the `VERCEL_ENV`/`VERCEL_URL` branch, and strips trailing slashes — across `dist/app` and `dist/pages`.

## Why

The custom Ory Elements flow pages render the login/registration/recovery/verification flows **same-origin**. On a **Vercel preview deployment**, the unpatched helper prefers the static `VERCEL_URL` over the request host, so the same-origin `/self-service/*` POSTs (and their Kratos flow cookies) target the deployment URL instead of the ephemeral preview origin the user is on — breaking the flow. The patch makes the visiting origin win.

## Scope / safety

- Patch is **inert unless the custom UI is enabled** (`NEXT_PUBLIC_ORY_CUSTOM_UI=true`), since only the gated flow pages call the patched helper. **Production (flag unset) is unaffected.**
- Diff is just the patch file + `patchedDependencies` registration + lockfile. No source changes.
- `bun install` applied cleanly; all unit tests pass (287/287).

## Rollout

For previews to actually route to the custom UI, this must land together with the terraform `enable_custom_auth_ui` flag and `NEXT_PUBLIC_ORY_CUSTOM_UI=true` + `NEXT_PUBLIC_ORY_SDK_URL` set on the Preview/Staging envs. Staging is a stable alias and doesn't strictly need the patch; **ephemeral previews do.**

Follow-up to #428.

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -18,7 +18,7 @@
         "@opentelemetry/sdk-metrics": "^2.0.1",
         "@opentelemetry/sdk-node": "^0.218.0",
         "@opentelemetry/semantic-conventions": "^1.36.0",
-        "@ory/client-fetch": "^1.22.37",
+        "@ory/client-fetch": "1.22.22",
         "@ory/elements-react": "^1.2.0",
         "@ory/nextjs": "^1.0.0-rc.1",
         "@radix-ui/react-avatar": "^1.1.4",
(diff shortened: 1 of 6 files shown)


huv1k:
Closing for now. Approach A (same-origin OAuth2 authorize via the @ory/nextjs proxy) hit Hydra's OAuth2 login-CSRF check at the login_verifier continuation (`request_forbidden: CSRF value from the token does not match the data store`) — proxying Hydra's /oauth2/* is outside what the @ory/nextjs proxy supports, and the Google OIDC callback transits the real Ory domain, splitting the CSRF cookie. Pivoting to the Kratos-only approach (#427) which avoids the Hydra browser-OAuth flow entirely. Branch kept.

PR #437, opened 2026-06-18
fix: Improve timestamp alignment in build logs

Kraci:
### Problem

Each build-log row shows a relative elapsed time (e.g. 3.6s) followed by the absolute time (e.g. 04:50:49.96 PM). The elapsed time has a variable width, which caused two visible bugs:

1. Truncation — once elapsed time hit two digits (10.2s), the extra character pushed the row past the fixed-width column and the absolute time got clipped to ….
2. Misalignment — the variable-width elapsed time shifted the absolute time to a different horizontal position on every row, so timestamps didn't line up vertically and were hard to scan.

On top of that, the elapsed value itself jumped around: formatDurationCompact emitted unpadded fields (2m 8s vs 2m 27s), so the seconds column wandered between single- and double-digit widths.

### Solution

Gave the timestamp a stable, fixed-width layout so the whole column lines up vertically and is easy to scan top-to-bottom. The elapsed time and the absolute time now each occupy a fixed column, durations are zero-padded so digits don't shift between rows, and the full timestamp always fits without truncation.

### DIFF
--- a/src/features/dashboard/build/logs-cells.tsx
+++ b/src/features/dashboard/build/logs-cells.tsx
@@ -7,7 +7,6 @@ import {
 } from '@/features/dashboard/common/log-cells'
 import { formatDurationCompact } from '@/lib/utils/formatting'
 import CopyButtonInline from '@/ui/copy-button-inline'
-import { Badge, type BadgeProps } from '@/ui/primitives/badge'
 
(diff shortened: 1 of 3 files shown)


Kraci:
@cla-bot check

tvi (approved):

PR #438, opened 2026-06-18
feat: auth flow via same-origin ory elements

drankou:
## Summary

Replaces the NextAuth (Auth.js) layer with a first-party, same-origin auth flow built directly on Ory. The dashboard now acts as Hydra's OAuth2 client **and** its consent/logout provider, drives an authorization-code + PKCE flow, and carries the resulting OIDC tokens in a single JWE-encrypted `e2b_session` cookie. Kratos owns the session; the cookie is just a token cache, never the auth gate.

This removes `next-auth` / `@auth/core` entirely and the Auth.js request-wrapping that had been fighting our sign-out and tRPC code paths.

## Why

- Auth.js wrapped every request and re-issued a JWT session cookie on the way out, which clobbered our sign-out cookie deletion and forced awkward boundary shims (`authjs-boundary`, `authjs-session-boundary`, `authjs-callbacks`).
- The OAuth2 redirect/PKCE/state handling was opaque inside the Auth.js provider, making the Hydra ⇄ Kratos consent/logout dance hard to reason about.

## What changed

**Auth flow (new routes)**
- `GET /api/auth/oauth/start` — builds the Hydra authorization request (PKCE S256, state, nonce), stashes the verifier/state/nonce in a short-lived httpOnly flow cookie, and 307-redirects to `/oauth2/auth`.
- `GET /api/auth/oauth/callback/ory` — validates state/nonce/PKCE, exchanges the code, bootstraps the dashboard user from the id_token, seals tokens into `e2b_session`. Failures route to the one-shot `recover` guard so a stale callback can't loop.
- `GET /consent` — dashboard is Hydra's consent provider; accepts the consent challenge and folds the identity's profile traits into the id_token (without this, tokens carry `sub`/`iss` but no email/name and bootstrap rejects the login).
- `GET /logout` — dashboard is Hydra's logout provider; accepts the logout challenge, routes through Kratos' own logout first so `ory_kratos_session` is cleared before Hydra finalizes.
- `sign-out` route now reads the id_token from `e2b_session` to build the Hydra RP-logout URL and clears the cookie on the redirect.

**Session / tokens**
- New `e2b_session` cookie: JWE (`dir` + `A256GCM`) via `jose`, carrying access/refresh/id tokens. Keyed by new `E2B_SESSION_SECRET` (replaces `AUTH_SECRET`); domain-scoped so it works across the cluster's subdomains.
- New `token-refresh` module refreshes the Hydra access token edge-side (direct token-endpoint call, refresh skew, dead/unchanged/refreshed states).
- New `kratos-session-edge` helper does an edge-safe Kratos `whoami` for the middleware redirect gate; authoritative enforcement stays in `getAuthContext`.

**UI**
- Ory Elements auth pages (`/login`, `/registration`, `/recovery`, `/verification`) are now always same-origin — the `NEXT_PUBLIC_ORY_CUSTOM_UI` flag and the legacy redirect fallbacks are gone.
- Auth footer links switched from `next/link` to plain `<a>` so the cross-origin start redirect stays a top-level document navigation (a soft nav would turn it into a CORS request Hydra rejects).

**Plumbing**
- tRPC handler no longer wrapped by `auth()`; auth is resolved per-procedure by the auth middleware (Kratos whoami + `e2b_session`).
- Deleted: `src/auth.ts`, the Auth.js boundary/callback shims, `next-auth.d.ts`, and the `[...nextauth]` / `bootstrap-failed` routes.

**Dependencies**
- Removed `next-auth` (+ `@auth/core` and its transitive `@panva/hkdf`, `preact*`).
- Added `jose` and `oauth4webapi` (direct, version-pinned).

**Env**
- Added `E2B_SESSION_SECRET`, optional `ORY_HYDRA_PUBLIC_URL` (self-hosted Hydra issuer).
- Removed `AUTH_SECRET`, `AUTH_TRUST_HOST`, `NEXT_PUBLIC_ORY_CUSTOM_UI`.

## Testing

New unit + integration coverage for the rewritten flow: `oauth-client`, `session-cookie`, `token-refresh`, `kratos-session-edge` (unit) and `auth-ory-callback`, `auth-ory-consent`, `auth-ory-logout`, plus the retained entrypoints/account-security suites (integration). The Auth.js session-boundary test was removed with its module.

## Notes for reviewers

- **Env / deploy:** set `E2B_SESSION_SECRET` (`openssl rand -hex 32`) in every environment before merge — `AUTH_SECRET` is no longer read.
- This is auth-critical; worth a careful pass on the callback failure/recover paths and the logout ordering (Kratos before Hydra finalize).


src/app/api/trpc/[trpc]/route.ts
Comment (ben-fornefeld) [inline src/app/api/trpc/[trpc]/route.ts:1]: the oryHandler existed for the following case: 
- token refresh happens one step before (inside the middleware)
- if the middleware refreshed the token, it appended a Set-Cookie, which resulted in procedures that have been called after the refresh but before the Set-Cookie reached the client, to use an already invalidated access token. 

does the new architecture account for this?

Comment (ben-fornefeld) [reply src/app/api/trpc/[trpc]/route.ts:1]: this was a known issue with authjs and they added helpers for this, which we were using. it might be that we don't have this case anymore now, just want to make sure we have this tracked

src/app/consent/route.ts
```diff
@@ -0,0 +1,102 @@
...
+}
+
+function fullName(
+  name: string | { first?: string; last?: string } | undefined
+): string | null {
+  if (!name) return null
+  if (typeof name === 'string') return name.trim() || null
+  const joined = [name.first, name.last].filter(Boolean).join(' ').trim()
+  return joined || null
+}
```
ben-fornefeld: do we need this? our identity schema specifies name as string afaik

src/app/login/components/custom-card.tsx
ben-fornefeld, line 1: do we need to replace Link with a? if the prefetch is the problem we can disabled it for these links

src/app/login/page.tsx
```diff
@@ -12,10 +9,6 @@ export const dynamic = 'force-dynamic'
 // getLoginFlow handles both legs of the OAuth2 login: a `login_challenge` from
 // Hydra (creates the Kratos flow) and the resulting `?flow=` from Kratos.
 export default async function OryLoginPage(props: OryPageParams) {
-  if (!isOryCustomUiEnabled()) {
-    redirect(AUTH_URLS.SIGN_IN)
-  }
```
ben-fornefeld, line 17: what is the migration path for this refactor?

src/core/server/auth/ory/identity.ts
```diff
@@ -1,22 +1,27 @@
...
-export function fromAuthSession(session: Session): AuthUser {
+// Build the user from a live Kratos session identity (whoami) — the source of
+// truth for getAuthContext. The session identity carries traits but not
+// credentials, so provider/credential flags stay false — use fromOryIdentity
+// with an admin lookup when those are needed (e.g. the profile query).
+export function fromKratosSessionIdentity(identity: {
+  id: string
+  traits?: unknown
+}): AuthUser {
+  const traits = (identity.traits ?? {}) as Record<string, unknown>
```
ben-fornefeld: should we add a zod.safeParse here for the traits we expect and if we can't parse, log an error?

the case here would be that ory identity schema has been changed and front-end does not know about it

src/app/api/trpc/[trpc]/route.ts
Comment (drankou) [reply src/app/api/trpc/[trpc]/route.ts:1]: yes, I was addressing this explicitly during the implementation. worth double check

src/core/server/auth/ory/oauth-flow.ts
ben-fornefeld, line 1: up to exploration: is there certain libraries we should consider using around the oauth state management ?

src/app/login/components/custom-card.tsx
drankou: yep, prefetch was the problem. we can disable it and use `<Link>`

src/core/server/auth/ory/session.ts
```diff
@@ -14,45 +14,55 @@ import type {
...
 
-export async function getAuthContext(
-  authSession?: Session | null
-): Promise<AuthContext | null> {
-  return getAuthContextFromOrySession(await readCurrentSession(authSession))
+// Kratos owns the session. Identity (the gate) comes from `whoami`; the Hydra
+// access token (API access only) comes from the e2b_session cookie, which the
+// middleware keeps fresh. For E2B the OAuth subject equals the Kratos identity
+// id, so kratos.identity.id is both the dashboard user id and the Kratos id used
+// for admin operations. This module never refreshes — it is a pure reader.
```
ben-fornefeld, line 44: the AuthUser.id we are using inside the dashboard state management should ideally map to E2B internal user id (public.users.id).

do we still account for that?

src/app/login/page.tsx
drankou: boolean flag in terraform, no FF on frontend

src/app/consent/route.ts
drankou: no need, fixed

src/core/server/auth/ory/identity.ts
drankou: added expected identity schema and zod parsing

src/core/server/auth/ory/oauth-flow.ts
drankou: we're using `oauth4webapi` lib for oauth flow, the rest is handled by ory's useFlow()

src/core/server/auth/ory/session.ts
drankou: we haven't matched it before afaik, and we shouldn't need this since kratos is source of truth for session, and API token has (iss, sub) pair included

src/app/consent/route.ts
```diff
@@ -0,0 +1,91 @@
...
+
+  const traits = (identity.traits ?? {}) as Partial<OryIdentityTraits>
+  const claims: Record<string, unknown> = {}
+
+  if (grantScope.includes('email') && traits.email) {
+    claims.email = traits.email
+  }
+
+  if (grantScope.includes('profile')) {
+    const name = traits.name?.trim()
```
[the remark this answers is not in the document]
drankou: name trait set as string in identity schema

src/app/api/auth/oauth/start/route.ts
```diff
@@ -1,34 +1,94 @@
...
+  } catch (error) {
+    l.error(
+      {
+        key: 'oauth_start:authorization_request_failed',
+        error: serializeErrorForLog(error),
+      },
+      'failed to build the Ory authorization request'
+    )
+    return NextResponse.redirect(new URL(AUTH_URLS.SIGN_IN, origin))
+  }
```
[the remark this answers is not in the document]
drankou: addressed

src/app/login/components/custom-card.tsx
```diff
@@ -14,14 +14,24 @@ export function OryCard({ children }: PropsWithChildren) {
...
   const { flowType } = useOryFlow()
 
   if (flowType === FlowType.Login) {
     return (
       <p className="text-fg-secondary mt-6">
         Don't have an account?{' '}
-        <Link href={AUTH_URLS.SIGN_UP} className="text-fg underline">
+        <Link
+          prefetch={false}
+          href={AUTH_URLS.SIGN_UP}
```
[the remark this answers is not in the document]
drankou: fixed

tests/integration/auth-ory-callback.test.ts
```diff
@@ -0,0 +1,127 @@
+import { NextRequest } from 'next/server'
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest'
+import {
+  E2B_OAUTH_FLOW_COOKIE,
+  serializeOryFlowState,
+} from '@/core/server/auth/ory/oauth-flow'
+import {
```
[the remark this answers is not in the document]
drankou: updated test

src/app/consent/route.ts
```diff
@@ -0,0 +1,91 @@
...
+    if (name) claims.name = name
+  }
+
+  return claims
```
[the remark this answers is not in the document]
drankou: fixed test

src/app/api/auth/oauth/callback/ory/route.ts
```diff
@@ -0,0 +1,111 @@
...
+      : null
+    return finalize(
+      NextResponse.redirect(logoutUrl ?? new URL(ORY_POST_LOGOUT_PATH, origin))
+    )
+  }
+
+  const sealed = await sealOrySession({
+    accessToken: tokens.accessToken,
+    refreshToken: tokens.refreshToken,
+    idToken: tokens.idToken,
```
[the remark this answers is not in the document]
drankou: fixed

src/app/consent/route.ts
```diff
@@ -0,0 +1,91 @@
...
+  try {
+    const oauth2 = getOryOAuth2Api()
+    const consent = await oauth2.getOAuth2ConsentRequest({ consentChallenge })
+    const grantScope = consent.requested_scope ?? []
+
+    const idTokenClaims = consent.subject
+      ? await profileClaimsForSubject(consent.subject, grantScope)
+      : {}
+
+    const { redirect_to } = await oauth2.acceptOAuth2ConsentRequest({
```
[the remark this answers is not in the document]
drankou: added check for client_id before consent

src/core/server/proxy/runtime.ts
```diff
@@ -47,30 +56,97 @@ function isOrySdkProxyPath(pathname: string): boolean {
...
-  if (isOrySdkProxyPath(request.nextUrl.pathname) && isOryCustomUiEnabled()) {
+  // classify as a bypass and go to Next).
+  if (isOrySdkProxyPath(request.nextUrl.pathname)) {
     return oryProxy(request)
   }
 
+  // Pattern B: refresh the e2b_session up front and propagate it to the same
+  // request (request.cookies) so RSC/route handlers and the gate below read the
+  // fresh token, then persist it on the outgoing response for the browser.
+  const session = await refreshSessionCookie(request)
```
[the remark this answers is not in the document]
drankou: addressed

src/core/server/auth/ory/kratos-session-edge.ts
```diff
@@ -22,8 +26,11 @@ export async function isKratosSessionActive(
       { headers: { cookie, accept: 'application/json' } }
     )
     if (!response.ok) return false
-    const session = (await response.json()) as { active?: boolean }
-    return session.active === true
+    const session = (await response.json()) as {
+      active?: boolean
+      identity?: { external_id?: string | null }
+    }
```
huv1k, line 32: Should we use Zod rather then type casting?

ben-fornefeld (approved):

PR #445, opened 2026-06-21
fix: fix truncated timestamps in sandbox logs and events and unify date formats

Kraci:
Consistent date/time/number formatting

### What changed

Pinned locale-sensitive formatting to en-US across the dashboard so timestamps, dates, and numbers render identically for every viewer (the timezone offset still follows the user's local zone). This keeps fixed-width columns stable and avoids SSR/client hydration mismatches. Affected surfaces: logs, events tab, sandboxes list, templates list/detail, sandbox header (started/ended), stopped banner, chart tooltips, monitoring memory marker, billing add-on dialog, and the calendar labels.

Widened timestamp columns to fit the full stamp and stop truncation — logs 164px → 190px, events tab 174px → 190px.

Time-range picker boundary messages now use the picker's own YYYY/MM/DD HH:MM:SS format instead of a browser-localized string, so "cannot be before/after …" messages match the input fields and can be typed straight back in.

Refactors

- Extracted a shared formatRelativeDateTime helper for the sandbox header (uses date-fns isToday/isYesterday), removing duplicated logic across started-at/ended-at.
- Logs timestamp cell now uses the shared formatLocalLogStyleTimestamp (with a -- fallback for invalid timestamps).
- Timezone resolution in the shared formatter is now opt-in (includeTimezone) to skip an expensive Intl call on hot virtualized rows.

| Before | After |
|--------|--------|
| Screenshot 2026-06-19 at 16 36 27 (1) | Screenshot 2026-06-21 at 18 31 30 |
| Screenshot 2026-06-19 at 18 18 29 (1) | Screenshot 2026-06-21 at 18 31 42 |

### DIFF
--- a/src/features/dashboard/billing/concurrent-sandboxes-addon-dialog.tsx
+++ b/src/features/dashboard/billing/concurrent-sandboxes-addon-dialog.tsx
@@ -142,11 +142,11 @@ function DialogContent_Inner({
   const limitIncreaseText = currentConcurrentSandboxesLimit ? (
     <>
       Increases total concurrent sandbox limit from{' '}
-      <b>{currentConcurrentSandboxesLimit.toLocaleString()}</b> to{' '}
+      <b>{currentConcurrentSandboxesLimit.toLocaleString('en-US')}</b> to{' '}
       <b>
-        {(
-          currentConcurrentSandboxesLimit + SANDBOXES_PER_ADDON
-        ).toLocaleString()}
+        {(currentConcurrentSandboxesLimit + SANDBOXES_PER_ADDON).toLocaleString(
+          'en-US'
+        )}
       </b>
     </>
   ) : (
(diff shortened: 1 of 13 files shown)


drankou:
Giving changes you made and issue's context, could you quickly check and comment on e2b-dev/dashboard#364 to make conflict resolution easier and avoid reintroducing similar bugs 🙏

huv1k (approved):

Kraci:
The only logic that was changed here is the use of unified US date and time formats. There are no changes to time zones. But yes, I’ll comment in this PR that I made that change. 👍

huv1k:
> The only logic that was changed here is the use of unified US date and time formats. There are no changes to time zones. But yes, I’ll comment in this PR that I made that change. 👍

I wonder if we should make it as default one across the board instead of putting it everywhere.

Kraci:
What do you have in mind? Shared utils for handling dates and times?

Kraci:
It would make sense. Let's do a follow up once we have timezones merged as well

PR #448, opened 2026-06-23
feat: paginate sandboxes list and cover paused + running states

huv1k:
## What

Reworks the sandboxes list page (`/dashboard/[teamSlug]/sandboxes/list`) to address [EN-1030](https://linear.app/e2b/issue/EN-1030/add-pagination-and-state-coverage-to-the-sandboxes-list):

- **State coverage** — migrate from the legacy running-only `GET /sandboxes` to `GET /v2/sandboxes` with `state=[running, paused]`, so the list now shows **both paused and running** sandboxes. A new **State** column renders a running/paused badge.
- **Pagination** — `getSandboxes` is now a cursor-based infinite query (`limit` + `nextToken`, reading the next cursor from the `X-Next-Token` response header). The table infinite-scrolls (prefetch-on-scroll + a Load more fallback), mirroring the existing templates list.
- **Drop the live-metrics overlay** — paused sandboxes have no live CPU/RAM/disk metrics, so the per-row metrics polling is removed. CPU/Memory/Disk columns now show **allocated specs**. The now-dead metrics code (`use-sandboxes-metrics`, `metrics-store`, the `getSandboxesMetrics` tRPC query + repository method, `transformMetricsToClientMetrics`, and the mock-metrics generator) is removed.

| Before | After |
| ------ | ----- |
| CleanShot 2026-06-23 at 16 11 58@2x | CleanShot 2026-06-23 at 16 12 07@2x |


src/features/dashboard/sandboxes/list/table.tsx
```diff
@@ -131,7 +152,7 @@ export default function SandboxesTable() {
 
   const table = useReactTable<SandboxListRow>({
     columns: sandboxListColumns,
-    data: data.sandboxes,
+    data: sandboxes,
```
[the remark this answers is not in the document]
huv1k: This will be done as followup later on

src/app/dashboard/[teamSlug]/sandboxes/(tabs)/list/page.tsx
```diff
@@ -11,6 +11,8 @@ export default async function SandboxesListPage({
   prefetch(
     trpc.sandboxes.getSandboxes.queryOptions({
       teamSlug,
+      limit: 50,
+      states: ['running'],
```
Comment (drankou) [inline src/app/dashboard/[teamSlug]/sandboxes/(tabs)/list/page.tsx]: shouldn't we try to prefetch both? or it's on purpose?

Comment (huv1k) [reply src/app/dashboard/[teamSlug]/sandboxes/(tabs)/list/page.tsx]: I removed it so it's same as right now

drankou (approved):

PR #461, opened 2026-06-24
[codex] Allow public templates in dashboard terminal

matthewlouisbrockman:
## What changed

- Allows `owner/template` references through the dashboard terminal template preflight.
- Keeps local/default template availability checks for bare template names.
- Adds unit coverage for public template reference detection.

## Why

`/sbx/new?template=owner/template` worked because it calls `Sandbox.create(template)` directly. The dashboard terminal page rejected the same public template reference before launch because it only checked default templates and templates owned by the current team.

This made public workshop links like `aiengineer-d56d/aiengineer-guide` show as unavailable in the terminal UI even though the SDK could create them.

## Validation

- `bunx vitest run tests/unit/dashboard-terminal.test.ts`
- `bunx biome check 'src/app/dashboard/[teamSlug]/terminal/page.tsx' src/features/dashboard/terminal/template.ts tests/unit/dashboard-terminal.test.ts`


matthewlouisbrockman:
Closing and replacing this with the correctly named Linear-task branch: overly-restrictive-template-checks-on-dashboardtabterminal-eng-4352.

PR #464, opened 2026-06-25
fix: improve sidebar hover line and add tooltip

drankou:
Improves the sidebar collapse affordance. The edge rail now shows a "Toggle Sidebar" tooltip with the keyboard shortcut. Consolidates the toggle shortcut on Cmd/Ctrl+B and keeps Ctrl+S binding for users with muscle memory.

| Before | After |
| ------- | ------- |
| CleanShot 2026-06-25 at 16 10 59@2x | CleanShot 2026-06-25 at 16 10 44@2x |

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/features/dashboard/sidebar/toggle.tsx
+++ b/src/features/dashboard/sidebar/toggle.tsx
@@ -15,6 +15,7 @@ export default function DashboardSidebarToggle() {
 
   const isOpen = open || openMobile
 
+  // Undocumented fallback so users used to the old Ctrl+S shortcut aren't confused.
   useKeydown((event) => {
     if (event.key === 's' && event.ctrlKey) {
       event.preventDefault()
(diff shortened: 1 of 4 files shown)


src/features/dashboard/sidebar/toggle.tsx
```diff
@@ -47,7 +39,7 @@ export default function DashboardSidebarToggle() {
           </motion.span>
         )}
       </AnimatePresence>
-      <ShortcutTooltip keys={['ctrl', 's']}>
+      <ShortcutTooltip keys={['cmd', 'b']}>
```
Kraci, line 51: Shouldn't we keep ctrl-s so we won't break the old shortcut? And show the new one in tooltip

Kraci (approved):

src/features/dashboard/sidebar/toggle.tsx
drankou: yes, addressed

PR #487, opened 2026-07-01
fix(posthog): identify users before first pageview via init bootstrap

huv1k:
## Problem

Investigation against PostHog project **E2B (33931)** showed ~17% of `/usage` visitors were recorded as anonymous (device-UUID `distinct_id`s, no email), because posthog-js autocaptures the first `$pageview` at `init()` before the deferred `posthog.identify()` runs — and person-on-events mode freezes those anonymous properties permanently. Separately, some identified persons had `email = null` written to their profile.

## Fix

`posthog.init()` now receives a `bootstrap` identity (the server-known user id via a new `PostHogBootstrap` component rendered in the dashboard layout), so the first captured pageview is already attributed to the real user instead of an anonymous device id; init is gated on that identity so it still never fires on marketing/auth routes. Since `bootstrap` only consumes identity/flag fields (not `$set` person properties), the email is set via `posthog.setPersonProperties()` inside the `loaded` callback — which posthog-js runs before it auto-captures the first `$pageview`, so that pageview is ingested with the email already on the person. The identity bridge also now sets the `email` property only when non-null.

Both fixes are forward-only — already-ingested anonymous events cannot be re-stamped via the SDK.

### DIFF
--- a/src/features/posthog-bootstrap.tsx
+++ b/src/features/posthog-bootstrap.tsx
@@ -1,0 +1,22 @@
+'use client'
+
+import { useEffect } from 'react'
+import { useAppPostHogProvider } from '@/features/posthog-provider'
+
+// Feeds the server-known identity into PostHog before init, so the first
+// captured pageview is already identified instead of anonymous.
+export function PostHogBootstrap({
+  distinctId,
+  email,
+}: {
+  distinctId: string
+  email?: string
+}) {
+  const { setBootstrap } = useAppPostHogProvider()
+
+  useEffect(() => {
+    setBootstrap({ distinctID: distinctId, email })
+  }, [distinctId, email, setBootstrap])
+
+  return null
+}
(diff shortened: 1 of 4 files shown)


src/features/posthog-bootstrap.tsx
```diff
@@ -0,0 +1,22 @@
+'use client'
+
+import { useEffect } from 'react'
+import { useAppPostHogProvider } from '@/features/posthog-provider'
+
+// Feeds the server-known identity into PostHog before init, so the first
+// captured pageview is already identified instead of anonymous.
+export function PostHogBootstrap({
```
drankou, line 8: do we need to have `OryPostHogIdentityBridge` as well?

huv1k: Yeah, we still need it — the two cover different things:

- **`PostHogBootstrap`** runs once at `init` from the *server* identity (`authContext.user`). It only fixes the initial `distinct_id`/`isIdentifiedID` + sets `email` before the first autocaptured `$pageview`. It does **not** do the `team` group.
- **`OryPostHogIdentityBridge`** is the only place that calls `posthog.group('team', …)` (grep-confirmed), and it re-identifies from the *client-refetched* user/team (`resolvedUser`/`initialTeam`), which can differ from the server snapshot — e.g. an email that was null at bootstrap but resolves later, or a team switch. On the happy path its `identify` is a no-op (same id + `setPersonProperties` dedupes), so the overlap is harmless.

So the bridge stays for the team group + keeping identity fresh on the client. I could fold the `group` call into a single component if you'd prefer fewer moving parts, but they have genuinely different lifecycles (init-time server data vs. reactive client data). Happy to consolidate if you want.

drankou: but `OryPostHogIdentityBridge` has .identify() call, so we can then unify into one solution? either existing bridge or the new Bootstrap?

PR #498, opened 2026-07-07
feat: split e2b_session cookie if too large

drankou:
The `e2b_session_` cookie carries the JWE-sealed OIDC tokens (accessToken + refreshToken + idToken). At ~3.7 KB it sits right under the browser's ~4096-byte per-cookie limit — which includes the name and all attributes. If a token grows (extra scopes/claims, longer refresh token), the cookie silently exceeds 4 KB and the browser drops it entirely → user appears logged out / access token vanishes mid-session.

## Solution
Split the sealed value across e2b_session.0, e2b_session.1, … past a safe threshold and reassemble on read — the same scheme authjs uses for authjs.session-token.

**Backward compatible**: small values stay on the bare e2b_session name; existing cookies keep working and re-chunk transparently on the next refresh.

* Small value → bare `e2b_session`. Large value → `e2b_session.0/.1/`….
* Every write reconciles the jar: writes the new chunks and expires stale ones (grow bare→chunked, shrink chunked→bare, or N→fewer chunks) so no orphan cookies linger.
* Read prefers numbered chunks (sorted numerically), falls back to the bare cookie, never mixes the two shapes.

fixes EN-1591

### DIFF
--- a/src/app/api/auth/oauth/callback/ory/route.ts
+++ b/src/app/api/auth/oauth/callback/ory/route.ts
@@ -12,9 +12,10 @@ import {
 import { resolveOryRedirectUri } from '@/core/server/auth/ory/oauth-relay'
 import { readKratosExternalId } from '@/core/server/auth/ory/session'
 import {
-  E2B_SESSION_COOKIE,
   ORY_SIGNUP_METADATA_COOKIE,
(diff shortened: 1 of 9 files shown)


Kraci (approved):
lgtm!

PR #514, opened 2026-07-16
Target feature flags by stable team ID

matthewlouisbrockman:
## Summary
- remove dashboard slugs from the shared feature-flag context
- stop forwarding `team.slug` to LaunchDarkly
- update every existing feature-flag call site and the provider contract test

## Why
Dashboard slugs are mutable routing identifiers, while the verified team ID is the stable LaunchDarkly targeting key. Keeping slug in the evaluation context allowed future rules to produce inconsistent results for the same team.

A read-only audit of the Dashboard LaunchDarkly project found no slug-based clauses or segments in Test, Staging, or Production. Existing team targets use team context keys with UUID team IDs, so this does not change current targeting behavior.

## Validation
- `bunx vitest run tests/unit/feature-flags.test.ts tests/unit/sidebar.test.ts`
- `bun run next typegen && bunx tsc --noEmit`
- `bun run knip:ci`
- focused Biome check
- `bunx react-doctor --verbose --diff origin/main --blocking warning --no-score --no-telemetry`

### DIFF
--- a/src/app/dashboard/[teamSlug]/agents/page.tsx
+++ b/src/app/dashboard/[teamSlug]/agents/page.tsx
@@ -41,7 +41,6 @@ export default async function AgentsPage({ params }: AgentsPageProps) {
     },
     team: {
       id: teamId.data,
-      slug: teamSlug,
     },
   })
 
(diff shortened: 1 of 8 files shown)


matthewlouisbrockman:
eh, will wait to coordinate this, realized someone else had added this not me

PR #518, opened 2026-07-23
refactor: fully self-hostable against e2b-dev/infra

ben-fornefeld:
This PR refocuses the dashboard on one goal: **managing a self-hosted [e2b-dev/infra](https://github.com/e2b-dev/infra) deployment with the least possible setup.** It also lays the groundwork for native OAuth support in the open-source stack: with the dashboard reduced to a stateless API-key client, identity can be added in front of it (or natively in infra) without reworking the dashboard again.

Until now, running this dashboard yourself meant standing up a whole constellation of services the open-source infra doesn't ship: an Ory Kratos + Hydra identity stack, a billing API backed by Stripe, LaunchDarkly, PostHog, Plain, and an events service. None of these exist in a self-hosted infra deployment — so out of the box, the dashboard simply couldn't be self-hosted without significant extra infrastructure that had nothing to do with managing sandboxes.

This PR strips the dashboard down to exactly what e2b-dev/infra supports, and replaces user-account auth with the credential every E2B deployment already has: a **team API key**.

## What this looks like now

**Setup is one env var:**

```bash
NEXT_PUBLIC_E2B_DOMAIN=your-domain.com   # resolves infra-api + dashboard-api
bun install && bun run dev
```

Visit `/`, paste a team API key (`e2b_...`), done. The key is validated against infra-api and stored in an httpOnly `e2b_api_key` cookie; every upstream call happens server-side via `X-API-Key`. The key never reaches client JavaScript.

For single-user deployments, set `E2B_API_KEY` in the environment and the dashboard is pre-authenticated — no login step at all.

**Authentication is deliberately out of scope.** The dashboard itself carries no user model. If you need SSO / user-level access control in front of it, put any standard **OAuth proxy** (oauth2-proxy, Authelia, Cloudflare Access, ...) in front of the deployment — the dashboard composes cleanly with that, instead of forcing a specific identity stack on you.

## What's included

- **Sandboxes** — live paginated list, per-sandbox monitoring (CPU/mem/disk), logs, filesystem inspector, in-browser terminal
- **Templates** — list, visibility, tags, build history with streaming build logs
- Teamless URLs: `/sandboxes`, `/templates/...` (the team is implied by the API key)
- `/?tab=sandboxes|templates` deep links

## What's removed (and why)

Everything that depends on services the open-source infra doesn't provide:

| Removed | Depended on |
|---|---|
| User accounts, login, CLI auth, account settings | Ory Kratos + Hydra |
| Billing, usage, limits pages | closed billing API + Stripe |
| Team & member management, API key management, team switcher | user-scoped auth-provider endpoints |
| Webhooks & sandbox event timeline | events service (argus) |
| Team-level metrics charts | team-scoped metrics endpoints not exposed via API key auth |
| Feature flags, support widget, analytics | LaunchDarkly, Plain, PostHog/GA/Vercel |
| Marketing site reverse proxy on `/` | e2b.dev landing page |

Net effect: **−48k lines, 20 fewer runtime dependencies**, and an env surface that went from ~25 required/conditional variables to **one**.

Where possible, code structure (repositories, tRPC routers, the envd data plane for the terminal/filesystem) was kept intact rather than rewritten, so the codebase stays familiar and easy to contribute to.

## Notable implementation details

- The proxy shrank to a ~30-line gate: cookie presence check + `returnTo`. Key validity is enforced lazily — any upstream 401 clears the cookie and returns to the key form.
- Sandbox terminal/filesystem still use the sandbox-scoped envd token data plane; the account-level credential never reaches the browser (unchanged design).
- `GET /sandboxes/{id}/record` (killed-sandbox archival records) isn't available on self-hosted deployments; the fallback path short-circuits to 404 while keeping the code shape.
- CSRF: `SameSite=Lax` httpOnly cookie + server-action origin checks; the `e2b_api_key` cookie is in the logger redaction list.

## Merge dependency

dashboard-api (from e2b-dev/infra) needs `X-API-Key` auth on its read endpoints (`GET /templates*`, `GET /builds*`) — tracked separately and landing before this merges. infra-api already accepts API key auth on every endpoint the dashboard uses. Until then, template/build detail views return 401 against a live deployment; sandboxes work fully.

## Testing

- Unit + integration suites rewritten for the new auth (proxy gate matrix, cookie/env fallback, key validation)
- Playwright preview tests are removed for now; end-to-end coverage will return once the self-hosted test setup is in place
- Verified against a production build: key form → sandboxes/templates → sign-out


tvi (approved):
Really excited about this direction. Decoupling the dashboard from identity is a key milestone on the path toward native OAuth support. It gives us the flexibility to support standard OAuth flows without coupling the UI to any particular identity provider or authentication stack.

PR #522, opened 2026-08-08
refactor(dashboard): replace xterm.js with rioterm

raphamorim:
This swaps the dashboard terminal from xterm.js to [rioterm](https://github.com/raphamorim/riotermjs), Rio terminal's Rust VT engine compiled to WebAssembly (full disclosure: I'm the author). The motivation is mostly deletion: `terminal-size.ts` measures cells by reading xterm's private `_core._renderService` and querying `.xterm-*` DOM nodes, which can break on any minor; rioterm exposes cell metrics publicly and `renderer.fit()` does the math, so that file and most of its tests go away. The WebGL context-loss/canvas-fallback dance becomes `renderer: 'canvas'` since parsing happens in wasm and the paint is a plain 2D canvas. Net ~-350 lines, hook contract unchanged (`dashboard-terminal.tsx` untouched), 243/243 unit tests pass, biome clean. Two honest notes: `cursorBlink` isn't supported yet (I'd add it upstream rather than fake it here), and I haven't run this against a live sandbox PTY, so a maintainer smoke test of the panel is the right bar before merging.

On performance, measured on byte-identical workloads against xterm.js with its WebGL addon: 1.9x faster cold init (21.6ms vs 41.1ms), 3.7x faster plain-text parse and paint (657 vs 179 MB/s), 2.2x faster on ANSI-heavy output (237 vs 109 MB/s), and up to 6.8x faster headless parsing, with both engines pinned at display refresh on full-screen TUI redraws. The suite and its caveats are in [benchmark/](https://github.com/raphamorim/riotermjs/tree/main/benchmark) if you want to rerun. The same adapter pattern runs in production for Lovable's sandbox terminal. Happy to iterate on whatever review surfaces.


raphamorim:
@cla-bot check

